PR #3, opened 2014-11-13
added single service and multiple service tests

mattyw:
Added some tests to the bakery for single service and multiple service scenarios

### DIFF
--- a/bakery/service_test.go
+++ b/bakery/service_test.go
@@ -1,0 +1,135 @@
+package bakery_test
+
+import (
+	"fmt"
+
+	gc "gopkg.in/check.v1"
+
+	"gopkg.in/macaroon-bakery.v0/bakery"
+	"gopkg.in/macaroon.v1"
+)
+
+type ServiceSuite struct{}
+
+var _ = gc.Suite(&ServiceSuite{})
+
+type strCompFirstPartyChecker string
+
+func (c strCompFirstPartyChecker) CheckFirstPartyCaveat(caveat string) error {
+	if caveat != string(c) {
+		return fmt.Errorf("%v doesn't match %s", caveat, c)
+	}
+	return nil
+}
+
+type strCompThirdPartyChecker string
+
+func (c strCompThirdPartyChecker) CheckThirdPartyCaveat(caveatId string, caveat string) ([]bakery.Caveat, error) {
+	if caveat != string(c) {
+		return nil, fmt.Errorf("%v doesn't match %s", caveat, c)
+	}
+	return nil, nil
+}
+
+// TestSingleServiceFirstParty creates a single service
+// with a macaroon with one first party caveat.
+// Created a request with this macaroon and checks that the service
+// can verify this macaroon as valid.
+func (s *ServiceSuite) TestSingleServiceFirstParty(c *gc.C) {
+	p := bakery.NewServiceParams{
+		Location: "loc",
+		Store:    nil,
+		Key:      nil,
+		Locator:  nil,
+	}
+	service, err := bakery.NewService(p)
+	c.Assert(err, gc.IsNil)
+
+	macaroon, err := service.NewMacaroon("", nil, nil)
+	c.Assert(err, gc.IsNil)
+	c.Assert(macaroon.Location(), gc.Equals, "loc")
+	cav := bakery.Caveat{
+		Location:  "",
+		Condition: "something",
+	}
+	err = service.AddCaveat(macaroon, cav)
+	c.Assert(err, gc.IsNil)
+
+	checker := strCompFirstPartyChecker("something")
+	req := service.NewRequest(checker)
+
+	req.AddClientMacaroon(macaroon)
+
+	err = req.Check()
+	c.Assert(err, gc.IsNil)
+}
+
+// TestMacaroonPaperFig6 implements an example flow as described in the macaroons paper:
+// http://theory.stanford.edu/~ataly/Papers/macaroons.pdf
+// There are three services, ts, fs, as:
+// ts is a storage service which has deligated authority to a forum service fs.
+// The forum service wants to require its users to be logged into to an authentication service as.
+//
+// The client obtains a macaroon from fs (minted by ts, with a third party caveat addressed to as).
+// The client obtains a discharge macaroon from as to satisfy this caveat.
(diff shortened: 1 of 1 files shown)


bakery/service_test.go
```diff
@@ -0,0 +1,135 @@
...
+func (c strCompThirdPartyChecker) CheckThirdPartyCaveat(caveatId string, caveat string) ([]bakery.Caveat, error) {
+	if caveat != string(c) {
+		return nil, fmt.Errorf("%v doesn't match %s", caveat, c)
+	}
+	return nil, nil
+}
+
+// TestSingleServiceFirstParty creates a single service
+// with a macaroon with one first party caveat.
+// Created a request with this macaroon and checks that the service
```
rogpeppe: s/Created/It creates/ ?

bakery/service_test.go
```diff
@@ -0,0 +1,135 @@
...
+	// ts creates a macaroon.
+	tsMacaroon, err := ts.NewMacaroon("", nil, nil)
+	c.Assert(err, gc.IsNil)
+
+	// ts somehow gets the macaroon to fs. fs adds a third party caveat to be discharged by as.
+	err = fs.AddCaveat(tsMacaroon, bakery.Caveat{Location: "as-loc", Condition: "user==bob"})
+	c.Assert(err, gc.IsNil)
+
+	// client asks for a discharge macaroon for each third party caveat.
+	// TODO (mattyw) why add the first party location
```
rogpeppe: The first party location is so that the third party can have some idea of whom it's adding first-party caveats for.

bakery/service_test.go
```diff
@@ -0,0 +1,135 @@
...
+	// TODO (mattyw) Why does the third party checker pass the id in the encoded form?
```
rogpeppe: I found it useful so that the caveat checker can return an error which contains the encoded caveat. See the idservice example for an example of how it can be used.

bakery/service_test.go
```diff
@@ -0,0 +1,135 @@
...
+
+	"gopkg.in/macaroon-bakery.v0/bakery"
+	"gopkg.in/macaroon.v1"
+)
+
+type ServiceSuite struct{}
+
+var _ = gc.Suite(&ServiceSuite{})
+
+type strCompFirstPartyChecker string
```
rogpeppe: I'd suggest putting these utilities at the end of the file, after all the tests.

bakery/service_test.go
```diff
@@ -0,0 +1,135 @@
...
+// http://theory.stanford.edu/~ataly/Papers/macaroons.pdf
+// There are three services, ts, fs, as:
+// ts is a storage service which has deligated authority to a forum service fs.
+// The forum service wants to require its users to be logged into to an authentication service as.
+//
+// The client obtains a macaroon from fs (minted by ts, with a third party caveat addressed to as).
+// The client obtains a discharge macaroon from as to satisfy this caveat.
+// The target service verifies the original macaroon it delegated to fs
+// No direct contact between as and ts is required
+func (s *ServiceSuite) TestMacaroonPaperFig6(c *gc.C) {
```
rogpeppe, line 59: Very nice test.

rogpeppe:
That's a great start, thanks! LGTM.

I think it would be really nice to have a version of one of the above tests with some fuzzing so that we can check that if any link in the chain is changed (for example, any signature is tampered with, or extra caveats added after binding) that the authorization fails. I'm not quite sure of the best way to phrase it though...

bakery/service_test.go
```diff
@@ -0,0 +1,173 @@
...
+		dm.Bind([]byte("tampered-signature"))
+		req.AddClientMacaroon(dm)
+	}
+
+	err = req.Check()
+	c.Assert(err, gc.ErrorMatches, "verification failed: signature mismatch after caveat verification")
+}
+
+func setUpFig6Services(c *gc.C) (*bakery.Service, *bakery.Service, *bakery.Service) {
+	fsKeyPair, err := bakery.GenerateKey()
```
rogpeppe: How about making a helper function:

```
func newService(c *gc.C, location string, locator bakery.PublicKeyLocatorMap) *bakery.Service {
    keyPair, err := bakery.GenerateKey()
    c.Assert(err, gc.IsNil)

    svc, err := bakery.NewService(bakery.NewServiceParams{
        Location: location,
        Key:      keyPair,
        Locator:  s.locator,
    })
    c.Assert(err, gc.IsNil)
    if locator != nil {
        locator[location] = &keyPair.Public
    }
    return svc
}
```

Then setUpFig6Services becomes somewhat redundant.
You can just do this directly in the tests:
     locator := make(bakery.PublicKeyLocatorMap)
     as := newService(c, "loc-as", locator)
     ts := newService(c, "loc-ts", locator)
     fs := newService(c, "loc-fs", locator)

which probably makes the tests a little bit more obvious than setUpFig6Services, and opens the door to more similar tests too.
I wondered about putting the locator into ServiceSuite and making newService a method on ServiceSuite that adds the service location to the suite. YMMV about that.

bakery/service_test.go
```diff
@@ -0,0 +1,173 @@
...
+
+	err = req.Check()
+	c.Assert(err, gc.IsNil)
+}
+
+func (s *ServiceSuite) TestMacaroonPaperFig6FailsWithoutDischarges(c *gc.C) {
+	ts, fs, _ := setUpFig6Services(c)
+	tsMacaroon := createMacaroonWithThirdPartyCaveat(c, ts, fs, bakery.Caveat{Location: "as-loc", Condition: "user==bob"})
+
+	req := ts.NewRequest(strCompFirstPartyChecker(""))
```
rogpeppe, line 106: I suggest keeping the same comments in the same places as in TestMacaroonPaperFig6, so that the relation between the code is easier to see.

rogpeppe:
LGTM with a couple of suggestions above. Thanks!

bakery/service_test.go
mattyw: @rogpeppe having this function is a good idea. The problem with putting locator into ServiceSuite means we'd need to add setuptest and teardown test calls to init and empty it, for now I'm happy with it being an argument to newService

rogpeppe:
LGTM

PR #6, opened 2014-12-08
An initial implementation of MongoDB storage for bakery.

alesstimec:


bakery/mgostorage/storage.go
```diff
@@ -0,0 +1,110 @@
...
+
+func logf(f string, a ...interface{}) {
+	if debug {
+		log.Printf(f, a...)
+	}
+}
+
+// NewMemStorage returns an implementation of Storage
+// that stores all items in MongoDB.
+func NewMgoStorage(s *mgo.Session, database, collection string) (*mgoStorage, error) {
```
rogpeppe: How about passing the collection in directly?
Then the error checking would not be necessary and the mgoStorage type could be simpler.

Also, I suggest New as a name (mgostorage.New seems like a reasonable constructor).

bakery/mgostorage/storage.go
```diff
@@ -0,0 +1,110 @@
...
+	logf("storage.Put[%q] %q", location, item)
+
+	i := storageItem{Location: location, Item: item}
+
+	collection, close := s.itemsCollection()
+	defer close()
+
+	_, err := collection.Upsert(bson.M{"location": location}, i)
+	if err != nil {
+		logf("storage.Put: error storing an item item: %s", err.Error())
```
rogpeppe: I think that if we're returning an error that we should be able to rely on the caller logging that error, so I'd lose the logfs here and below.

bakery/mgostorage/storage.go
```diff
@@ -0,0 +1,110 @@
...
+	}, nil
+}
+
+type mgoStorage struct {
+	session *mgo.Session
+	db      string
+	c       string
+}
+
+type storageItem struct {
```
rogpeppe: This collection may be quite big, so how about using "loc" rather than location
for the field name and save 5 bytes per entry without too much loss of readability?

Also, I'd use storageDoc for the type name - it's slightly more conventional for a mongo doc
type.

bakery/mgostorage/storage.go
```diff
@@ -0,0 +1,110 @@
...
+
+func (s *mgoStorage) setupCollection() error {
+	collection, close := s.itemsCollection()
+	defer close()
+
+	return collection.EnsureIndex(mgo.Index{Key: []string{"location"}, Unique: true})
+}
+
+// itemsCollection returns the items collection
+func (m *mgoStorage) itemsCollection() (*mgo.Collection, func()) {
```
rogpeppe: How about something like this instead?

```
// collection returns the collection with a copied mgo session.
// It must be closed when done with.
func (m *mgoStorage) collection() collectionCloser {
     c := s.coll.Database.Session.Copy().DB(s.coll.Database.Name).C(s.coll.C)
     return collectionCloser{c}
}

type collectionCloser struct {
      *mgo.Collection
}
func (c collectionCloser) Close() {
     c.Collection.Database.Session.Close()
}
```

Then the various callers would be written:

```
 collection := s.collection()
 defer collection.Close()
```

and we produce no more garbage than necessary.

bakery/mgostorage/storage.go
```diff
@@ -0,0 +1,110 @@
...
+// Put implements the bakery Storage interface.
+func (s mgoStorage) Put(location, item string) error {
+	logf("storage.Put[%q] %q", location, item)
+
+	i := storageItem{Location: location, Item: item}
+
+	collection, close := s.itemsCollection()
+	defer close()
+
+	_, err := collection.Upsert(bson.M{"location": location}, i)
```
rogpeppe: I'd use bson.D here and below - it's quite a bit more efficient and almost as readable.

bakery/mgostorage/storage.go
```diff
@@ -0,0 +1,110 @@
...
+	if err != nil {
+		logf("storage.Put: error storing an item item: %s", err.Error())
+		return errors.New("storage.Put : error storing an item")
+	}
+	return nil
+}
+
+// Get implements the bakery Storage interface.
+func (s mgoStorage) Get(location string) (string, error) {
+	logf("storage.Get[%q]", location)
```
rogpeppe: I think these informational logfs are going to end up way too verbose. I'd leave them out, at least for the time being.

bakery/mgostorage/storage.go
```diff
@@ -0,0 +1,110 @@
...
+	defer close()
+
+	var i storageItem
+	err := collection.Find(bson.M{"location": location}).One(&i)
+	if err != nil {
+		if err == mgo.ErrNotFound {
+			return "", bakery.ErrNotFound
+		}
+		logf("storage.Get[%q] -> error finding an item: %v", location, err)
+		return "", errors.New("error finding an item for location " + location)
```
rogpeppe: We want to include the error here please. The bakery uses errgo, so I'd suggest:

```
 return "", errgo.Notef(err, "cannot get %q", location)
```

bakery/mgostorage/storage.go
```diff
@@ -0,0 +1,110 @@
...
+// Del implements the bakery Storage interface.
+func (s mgoStorage) Del(location string) error {
+	logf("storage.Del[%q]", location)
+	collection, close := s.itemsCollection()
+	defer close()
+
+	err := collection.Remove(bson.M{"location": location})
+	if err != nil {
+		logf("storage.Del[%q] -> error removing an item: %v", location, err)
+		return errors.New("error removing an item for location " + location)
```
rogpeppe: Similar errgo suggestion here.

bakery/mgostorage/storage_integration_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+	"github.com/juju/testing"
+	gc "gopkg.in/check.v1"
+	"gopkg.in/mgo.v2"
+
+	"github.com/go-macaroon-bakery/macaroon-bakery/bakery"
+	storage "github.com/go-macaroon-bakery/macaroon-bakery/bakery/mgostorage"
+)
+
+type StorageIntegrationSuite struct {
```
rogpeppe: Let's just call this StorageSuite.

rogpeppe: Ah, I see there already is a StorageSuite. For one test, I'm not sure this suite is worth it. I think I'd just put all of the bakery stuff into a single TestCreateMacaroon test on StorageSuite, and factor it out latest if it becomes necessary.

rogpeppe:
Thanks a lot for this. It looks great, with a bunch of fairly trivial suggestions.

bakery/mgostorage/storage.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/go-macaroon-bakery/macaroon-bakery/bakery"
+)
+
+// New returns an implementation of Storage
+// that stores all items in MongoDB.
+func New(c *mgo.Collection) *mgoStorage {
+	m := mgoStorage{
+		col: c,
+	}
+	m.setupCollection()
```
rogpeppe: Now setupCollection is being called, we'll need to return an error from New,
otherwise any error is always ignored.

alesstimec: Yes, i just noticed that, too.. Fix is on the way.

bakery/mgostorage/storage.go
```diff
@@ -0,0 +1,96 @@
...
+type mgoStorage struct {
+	col *mgo.Collection
+}
+
+type storageDoc struct {
+	Location string `bson:"loc"`
+	Item     string `bson:"item"`
+}
+
+func (s *mgoStorage) setupCollection() error {
```
rogpeppe: Apologies for this totally trivial nit: "setup" is a noun. "set up" is a verb. So this should technically be "setUpCollection".

bakery/mgostorage/storage.go
```diff
@@ -0,0 +1,96 @@
...
+
+type collectionCloser struct {
+	*mgo.Collection
+}
+
+func (c collectionCloser) Close() {
+	c.Collection.Database.Session.Close()
+}
+
+// Put implements the bakery Storage interface.
```
rogpeppe: I've been trying to use a standard form for these comments. I'd suggest:

  // Put implements bakery.Storage.Put.

It's slightly more accurate, as this method doesn't by itself implement the interface.

rogpeppe:
LGTM modulo returning an error from New and a couple of ultra-trivials. Looks cleaner now, I hope you'll agree. Thanks again.

PR #8, opened 2014-12-08
bakery: implement key marshaling

rogpeppe:
We change the types of keys so that they can both use the same
marshaling and unmarshaling code.

This is an API-breaking change.


jrwren:
:+1:

cmars:
LGTM

PR #10, opened 2014-12-09
bakery/mgostorage: do not return private type

rogpeppe:
It's not good practice to return private types.


mhilton:
:+1: LGTM

alesstimec:
LGTM

PR #11, opened 2014-12-09
httpbakery: expose slightly more errors machinery

rogpeppe:
We want to be able to integrate httpbakery into an existing JSON
framework.


cmars:
I'm not as familiar with jsonhttp as I should be, but this LGTM.

mhilton:
:+1: LGTM

PR #12, opened 2014-12-10
bakery: remove Request type.

rogpeppe:
httpbakery: remove Service type. Add cookie helpers.

Also change bakery.DischargeAll so that it returns a slice suitable
for putting in a cookie.


httpbakery/client.go
```diff
@@ -19,6 +19,36 @@ import (
 	"gopkg.in/macaroon-bakery.v0/bakery"
 )
 
+// DefaultHTTPClient is an http.Client that ensures that
```
mhilton, line 22: Is it a go bug that the default client doesn't do this?

httpbakery/client.go
```diff
@@ -110,48 +140,57 @@ func (ctxt *clientContext) do1(req *http.Request) (*http.Response, error) {
...
+	jar.SetCookies(url, []*http.Cookie{cookie})
+	return nil
+}
+
+func (ctxt *clientContext) addCookie(req *http.Request, ms []*macaroon.Macaroon) error {
+	cookies, err := NewCookie(ms)
+	if err != nil {
+		return errgo.Mask(err)
+	}
+	// TODO should we set it for the URL only, or the host.
```
mhilton, line 190: My reading of the cookie jar code is that it does only set it for the host portion. You can expand things further using the Domain value in the Cookie.

httpbakery/client.go
rogpeppe, line 22: Yup.
https://github.com/golang/go/issues/4677

httpbakery/client.go
rogpeppe, line 190: Cool. I'll remove the TODO then.

httpbakery/client.go
```diff
@@ -270,3 +309,66 @@ func postFormJSON(url string, vals url.Values, resp interface{}, postForm func(u
...
+		}
+	}
+	var err error
+	for _, ms := range mss {
+		err = svc.Check(ms, checker)
+		if err == nil {
+			return nil
+		}
+	}
+	// Return an arbitrary error from the macaroons provided.
```
mhilton, line 360: I feel this should be trying to find the most relevant macaroon to report the failure on, by some measure of relevance. However I suspect that is impossible.

mhilton:
:+1: LGTM

httpbakery/client.go
rogpeppe, line 360: Yeah. Another possibility would be to return all the errors, I guess.

bakery/discharge.go
```diff
@@ -33,6 +37,7 @@ func DischargeAll(
 		if err != nil {
 			return nil, errgo.NoteMask(err, fmt.Sprintf("cannot get discharge from %q", cav.Location), errgo.Any)
 		}
+		dm.Bind(sig)
```
cmars, line 40: I can't think of a reason you _wouldn't_ want to bind when discharging all the macaroons, but this behavior should at least be documented in the godoc comment.

Would binding twice break things? I have a feeling it would, I'd have to go look at the impl of Macaroon.Bind.

bakery/example/authservice.go
```diff
@@ -11,7 +11,7 @@ import (
 // that can discharge third-party caveats added
 // to other macaroons.
 func authService(endpoint string, key *bakery.KeyPair) (http.Handler, error) {
-	svc, err := httpbakery.NewService(bakery.NewServiceParams{
+	svc, err := bakery.NewService(bakery.NewServiceParams{
```
cmars, line 14: I really like the consolidation of httpbakery into bakery.

httpbakery/client.go
cmars, line 360: If you had multiple primary macaroons, each primary macaroon might fail to Check() for missing discharges. In that case, the caller needs to be able to react for each one that is retry-able with additional discharges. Something like (and this is a hand-wavy example):
`type CheckRequestError []error`
with a helper like:

```
// NeedsDischarge returns a slice containing all macaroons that require a discharge
// in the given error.
func NeedsDischarge(err error) []*macaroon.Macaroon
```

or, is this better handled a different way?

cmars, line 360: I'd like to see all the errors returned for now, even if we don't decide how to deal with them yet. Reminds me of the bulk API call pattern we like to use in Juju.

cmars:
LGTM with the bulk errors in CheckRequest addressed here or in a followup. @alesstimec note the API changes in this PR.

Thanks!

alesstimec:
@cmars Noted. Will change, as soon as this PR lands.

httpbakery/client.go
rogpeppe, line 360: I hope you don't mind, but I think it's best to leave this change for another PR. This one's already quite big, and making that change is not insignificant. I'm not sure that the situation is as bad as you paint it, because I don't think it will be useful for clients to act differently depending on whether macaroons have their discharges attached or not. It's the client's responsibility to add discharges - they can tell before sending them whether they need discharges. The verification logic can fail in all kinds of ways, and the intention is to respond to any of them by creating a new macaroon to be discharged. So the main reason for returning all errors is for informational purposes only AFAICS.

bakery/discharge.go
rogpeppe, line 40: It is already documented in the godoc doc: "All the discharge macaroons 
will be bound to the primary macaroon."

Yes, binding twice would break things. But I don't _think_ that should
be an issue - you can't usefully add caveats to a macaroon after
discharging it, so I think that doing the binding here makes sense.
If it turns out it's actually useful to be able to discharge but not bind,
we can always create a DischargeAllUnbound function in the future.

I'm hoping that no-one will need to call Bind directly any more.

httpbakery/client.go
rogpeppe, line 22: Added a link to the bug.

httpbakery/client.go
rogpeppe, line 190: Updated the TODO.

httpbakery/client.go
rogpeppe, line 360: Added a TODO.

PR #13, opened 2014-12-11
checkers: new Checker interface, add tests

rogpeppe:
This makes the checkers package actually useful.

### DIFF
--- a/bakery/checkers/checkers.go
+++ b/bakery/checkers/checkers.go
@@ -1,93 +1,172 @@
-// The checkers package provides some standard caveat
-// checkers and checker-combining functions.
+// The checkers package provides some standard first-party
+// caveat checkers and some primitives for combining them.
 package checkers
 
 import (
 	"fmt"
 	"strings"
-	"time"
+
+	"gopkg.in/errgo.v1"
 
 	"gopkg.in/macaroon-bakery.v0/bakery"
 )
 
-func FirstParty(condition string) bakery.Caveat {
-	return bakery.Caveat{
-		Condition: condition,
+// Checker is implemented by types that can check caveats.
+type Checker interface {
+	// Condition returns the identifier of the condition
+	// to be checked - the Check method will be used
+	// to check caveats with this identifier.
+	//
+	// It may return an empty string, in which case
+	// it will be used to check any condition
+	Condition() string
+
+	// Check checks that the given caveat holds true.
+	// The condition and arg are as returned
+	// from ParseCaveat.
+	//
+	// For a checker with an empty condition, a
+	// return of bakery.ErrCaveatNotRecognised from
+	// this method indicates that the condition was
+	// not recognized.
+	Check(cond, arg string) error
+}
+
+// New returns a new MultiChecker that uses all the
+// provided Checkers to check caveats. If several checkers return the
+// same condition identifier, all of them will be used.
+//
+// The cause of any error returned by a checker will be preserved.
+//
+// Note that because the returned checker implements Checker
+// as well as bakery.FirstPartyChecker, calls to New can be nested.
+// For example, a checker can be easily added to an existing
+// MultiChecker, by doing:
+//
+//	checker := checkers.New(old, another)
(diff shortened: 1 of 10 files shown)


bakery/checkers/checkers.go
```diff
@@ -1,93 +1,167 @@
...
-		err := c0.CheckFirstPartyCaveat(caveat)
-		if _, ok := err.(*bakery.CaveatNotRecognizedError); ok {
-			err = c1.CheckFirstPartyCaveat(caveat)
-		}
-		return err
+// CheckerFunc implements Checker for a function.
+type CheckerFunc struct {
+	// Condition_ holds the condition that the checker
+	// implements.
+	Condition_ string
```
mhilton, line 117: I'm not sure about ending variable names with underscores. I can see myself getting that wrong a lot.

How about Cond & CheckFunc?

bakery/checkers/checkers.go
```diff
@@ -1,93 +1,167 @@
...
+	checked := false
+	for _, c := range c.checkers {
+		checkerCond := c.Condition()
+		if checkerCond != "" && checkerCond != cond {
+			continue
+		}
+		if err := c.Check(cond, arg); err != nil {
+			if checkerCond != "" || errgo.Cause(err) != bakery.ErrCaveatNotRecognized {
+				return errgo.Mask(err, errgo.Any)
+			}
```
frankban, line 74: This could be written with a continue in the case of an error and without the need of an else block [shrug].
In the error case, when we don't return, a comment would probably help the reader.

bakery/checkers/checkers.go
```diff
@@ -1,93 +1,167 @@
...
-		Condition: condition,
-	}
+// Checker is implemented by types that can check caveats.
+type Checker interface {
+	// Condition returns the identifier of the condition
+	// to be checked - the Check method will be used
+	// to check caveats with this identifier.
+	//
+	// It may return an empty string, in which case
+	// it will be used to check any condition.
```
mhilton: Perhaps note here that ErrCaveatNotRecognized should be ignored if the checker returns ""?

bakery/checkers/checkers.go
```diff
@@ -1,93 +1,167 @@
...
+// ParseCaveat parses a caveat into an identifier, identifying the
+// checker that should be used, and the argument to the checker (the
+// rest of the string).
 //
-// The identifier is taken from all the characters
-// before the first space character.
-func ParseCaveat(cav string) (string, string, error) {
+// The identifier is taken from all the characters before the first
+// space character.
+func ParseCaveat(cav string) (cond, arg string, err error) {
```
mhilton, line 164: Thanks, that makes things clearer.

mhilton:
:+1: LGTM, with comments.

bakery/checkers/checkers.go
rogpeppe, line 117: I swithered about this. I started with something like your
suggestion, but in the end I decided that a strict
correspondence with the method names was actually more
predictable, and it potentially sets a precedent for this 
not-unusual case. It means that every time this happens,
we don't have to think of some (inevitably inconsistent) name
for the field that clashes with the method with the same name.

We have already used this idiom in other places (in errgo and the
charm tests and I see it's used by other people too (e.g. Russ Cox in
his FUSE code). So I'd prefer to leave it for now. If you find that
you're constantly getting it wrong, then we'll change it.

frankban:
:+1: Thank you!

bakery/checkers/checkers.go
rogpeppe: Done

bakery/checkers/checkers.go
rogpeppe, line 74: Good point. I got this logic wrong initially and it's still unclear.

Done, slightly differently, without so many negatives. Hopefully the new logic is now clear enough that the comment isn't needed, as it corresponds closely to the wording in the Checker.Check doc comment.

PR #14, opened 2014-12-11
bakery/checkers: add Declared and InferDeclared

rogpeppe:
This allows macaroons to declare information by adding a first-party
caveat. The declarations inferred are valid after caveat checking because
we guarantee to check all discharge macaroons when verifying. A client can
add another declaration caveat, but they can't change the value of an existing
declaration.

To come: built in support for third party caveats that require a set of
declarations to be added when discharged.

### DIFF
--- a/bakery/checkers/checkers_test.go
+++ b/bakery/checkers/checkers_test.go
@@ -4,8 +4,10 @@ import (
 	"fmt"
 	"time"
 
+	jc "github.com/juju/testing/checkers"
 	gc "gopkg.in/check.v1"
 	"gopkg.in/errgo.v1"
+	"gopkg.in/macaroon.v1"
 
 	"gopkg.in/macaroon-bakery.v0/bakery"
 	"gopkg.in/macaroon-bakery.v0/bakery/checkers"
@@ -176,6 +178,39 @@ var checkerTests = []struct {
 		caveat:      checkers.TimeBeforeCaveat(now).Condition + " ",
 		expectError: `caveat "time-before 2006-01-02T15:04:05.123Z " not fulfilled: parsing time "2006-01-02T15:04:05.123Z ": extra text:  `,
 	}},
+}, {
+	about:   "declared, no entries",
+	checker: checkers.New(checkers.Declared{}),
+	checks: []checkTest{{
+		caveat:      checkers.DeclaredCaveat("a", "aval").Condition,
+		expectError: `caveat "declared a aval" not fulfilled: got a=null, expected "aval"`,
+	}, {
+		caveat:      checkers.CondDeclared,
+		expectError: `caveat "declared" not fulfilled: declared caveat has no value`,
+	}},
+}, {
+	about: "declared, some entries",
+	checker: checkers.New(checkers.Declared{
+		"a":   "aval",
+		"b":   "bval",
+		"spc": " a b",
+	}),
+	checks: []checkTest{{
+		caveat: checkers.DeclaredCaveat("a", "aval").Condition,
+	}, {
+		caveat: checkers.DeclaredCaveat("b", "bval").Condition,
+	}, {
+		caveat: checkers.DeclaredCaveat("spc", " a b").Condition,
+	}, {
+		caveat:      checkers.DeclaredCaveat("a", "bval").Condition,
+		expectError: `caveat "declared a bval" not fulfilled: got a="aval", expected "bval"`,
+	}, {
+		caveat:      checkers.DeclaredCaveat("a", " aval").Condition,
+		expectError: `caveat "declared a  aval" not fulfilled: got a="aval", expected " aval"`,
+	}, {
+		caveat:      checkers.DeclaredCaveat("spc", "a b").Condition,
+		expectError: `caveat "declared spc a b" not fulfilled: got spc=" a b", expected "a b"`,
+	}},
 }}
 
 var errWrongArg = errgo.New("wrong arg")
@@ -195,6 +230,16 @@ func argChecker(expectCond, checkArg string) checkers.Checker {
 	}
 }
 
+func (s *CheckersSuite) TestDeclaredCaveatPanic(c *gc.C) {
+	c.Assert(func() {
+		checkers.DeclaredCaveat("", "hello")
+	}, gc.PanicMatches, `invalid caveat declared key ""`)
+
+	c.Assert(func() {
+		checkers.DeclaredCaveat("some thing", "hello")
+	}, gc.PanicMatches, `invalid caveat declared key "some thing"`)
+}
+
 func (s *CheckersSuite) TestMultiChecker(c *gc.C) {
 	c.Logf("time is %s", now)
 	for i, test := range checkerTests {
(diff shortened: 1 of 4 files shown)


bakery/checkers/checkers_test.go
```diff
@@ -214,3 +259,130 @@ func (s *CheckersSuite) TestMultiChecker(c *gc.C) {
...
+	expect: checkers.Declared{
+		"c": "cval",
+		"d": "dval",
+	},
+}, {
+	about: "third party caveats ignored",
+	caveats: [][]bakery.Caveat{{{
+		Condition: "declared a no conflict",
+		Location:  "location",
+	},
```
mhilton, line 352: This seems indented incorrectly

mhilton:
:+1: Looks good, I think it does need a lot of explanation about how to use it somewhere though.

bakery/checkers/checkers_test.go
rogpeppe, line 352: gofmt doesn't lie :)

bakery/checkers/declared.go
```diff
@@ -0,0 +1,94 @@
...
+// DeclaredCaveat returns a "declared" caveat asserting that the given key is
+// set to the given value. If a macaroon has exactly one first party
+// caveat asserting the value of a particular key, then InferDeclared
+// will be able to infer the value, and then DeclaredChecker will allow
+// the declared value.
+//
+// DeclaredCaveat will panic if the key is empty or contains a space character.
+func DeclaredCaveat(key string, value string) bakery.Caveat {
+	if strings.Contains(key, " ") || key == "" {
+		panic(fmt.Errorf("invalid caveat declared key %q", key))
```
urosj, line 22: Must we really panic here?

bakery/checkers/declared.go
```diff
@@ -0,0 +1,94 @@
...
+// check that any "declared" caveats have a matching
+// key for their value in the map.
+type Declared map[string]string
+
+// Condition implements Checker.Condition.
+func (c Declared) Condition() string {
+	return "declared"
+}
+
+// Check implements Checker.Check
```
urosj: A little more description here, please, as this bears no new information above the function name. At least what it does to data, mostly for documentation purposes as the code is self-explanatory. And add a "." at the end, please.

bakery/checkers/declared.go
```diff
@@ -0,0 +1,94 @@
...
+// InferDeclared retrieves any declared information from
+// the given macaroons and returns it as a key-value map.
+//
+// Information is declared with a first party caveat as created
+// by DeclaredCaveat.
+//
+// If there are two caveats that declare the same key with
+// different values, the information is omitted from the map.
+// When the caveats are later checked, this will cause the
+// check to fail.
```
urosj, line 66: Sounds like a good enough resolution. I wonder if there is a situation where this behaviour would be wrong. it would be nice to have a proof that this this false negative never happens.

bakery/checkers/declared.go
```diff
@@ -0,0 +1,94 @@
...
+	"gopkg.in/macaroon.v1"
+
+	"gopkg.in/macaroon-bakery.v0/bakery"
+)
+
+// DeclaredCaveat returns a "declared" caveat asserting that the given key is
+// set to the given value. If a macaroon has exactly one first party
+// caveat asserting the value of a particular key, then InferDeclared
+// will be able to infer the value, and then DeclaredChecker will allow
+// the declared value.
```
mattyw: If my understanding of this scheme is right, then I'd considering adding:

"and then DeclaredChecker will allow the declared value" if it matches the value provided here

bakery/checkers/declared.go
mattyw: In the comments I'd state explicitly why condition is not needed as an argument in this call

bakery/checkers/checkers_test.go
```diff
@@ -4,8 +4,10 @@ import (
 	"fmt"
 	"time"
 
+	jc "github.com/juju/testing/checkers"
```
mattyw, line 7: It seems a shame to require this package just for a single call to jc.DeepEquals

bakery/checkers/declared.go
```diff
@@ -0,0 +1,94 @@
...
+//
+// DeclaredCaveat will panic if the key is empty or contains a space character.
+func DeclaredCaveat(key string, value string) bakery.Caveat {
+	if strings.Contains(key, " ") || key == "" {
+		panic(fmt.Errorf("invalid caveat declared key %q", key))
+	}
+	return firstParty("declared", key+" "+value)
```
mattyw: I'm not sure if it's intended behaviour, but it could be possible to a user of this library to created a declared caveat without knowing they exist.

Let's say a user doesn't know about declared caveats and has decided that they want to mint a macaroon - but they've designed it with a caveat that uses declared as a condition:

```
m, err := svc.NewMacaroon("id", rootKey,[]Caveat{{"declared", "user-bob"}}) 
```

Is that likely to cause any problems.

Put simply - aren't we saying that "declared" is a keyword reserved for declared caveats - in that case do we want to choose a more unique name - and maybe have this defined as a const?

rogpeppe: This is an issue with any caveat - first party caveats will all be interpreted in the namespace of the first party. If someone adds a caveat that they intend to mean one thing, but the checker inteprets it a different way, there's little that can be done. That's why I want to standardise on checker names. We could perhaps have a convention for caveats that the first party defines locally, an "x-" or "local-" prefix for example. Third parties would never add caveats with that prefix.

bakery/checkers/checkers_test.go
rogpeppe, line 7: Yeah, but jc.DeepEquals is useful enough that I think it's worth it. I might consider proposing adding it to gocheck itself in the future.

bakery/checkers/declared.go
rogpeppe, line 66: Do you mean a false positive here?

My informal proof that this is OK goes something like:
- We know that all caveats are checked.
- If two caveats state different values for the same key, then the check will fail, because the same key cannot have two values.

So regardless of the value in the map, the check must fail if a key is declared with two different values. Therefore the behaviour of this function cannot cause a check to succeed where it would otherwise have failed.

The important thing is that if the check succeeds, the contents of this map hold the declared information.

bakery/checkers/declared.go
rogpeppe, line 22: The way this is intended to be used, I think it's OK.
In general, we will be declaring known keys. This
function is so simple that to make everyone check the
error when they're passing in a known constant seems
like overkill.

bakery/checkers/declared.go
mattyw: @rogpeppe is it at least worth adding "declared" as a const - with a comment that explains what it's for- Might be a useful way to highlight this feature in godoc

bakery/checkers/declared.go
rogpeppe: Done.

bakery/checkers/declared.go
rogpeppe: Done both.

bakery/checkers/declared.go
rogpeppe: Good idea. I'll add constants for all our conditions.

bakery/checkers/declared.go
```diff
@@ -0,0 +1,98 @@
+package checkers
```
mattyw, line 1: Do you think this file belongs here or in bakery/checkers/declared/declared.go?

mattyw:
@rogpeppe One comment regarding the package structure. In theory LGTM but I'd be interested in some feedback regarding the validity of this approach before it lands

bakery/checkers/declared.go
rogpeppe, line 1: I'm keeping all the standard checkers together for now.

cmars:
LGTM ~~with a followup on panic handling~~ . I think the approach deserves further security analysis, but this is the first step in getting there. Exciting, interesting stuff!

bakery/checkers/declared.go
rogpeppe, line 22: This can't be caused by an invalid macaroon. It can only be caused by someone _creating_ a bad caveat, which is a serious bug and one that's really not likely to happen. I reserve my right to change my mind about this when we find real code that triggers the panic, but for the time being, let's keep it trivial to create these caveats.

Another approach might be to log the error and just create a caveat which is always false.

urosj, line 22: I like the last proposal. You don't have to remove the panic in this PR, but I'd suggest to add a todo with " log the error and create a caveat which is always false" (skipped the "just" part :D).

bakery/checkers/declared.go
urosj, line 66: "If two caveats state different values for the same key, then the check will fail, because the same key cannot have two values." 

But we're coping with a situation of having to evaluate two different values for the same key :) Not saying that this will happen, but I can see the following scenario: you get different values for the same key, therefore invalidating further flow. The client retries, the same thing happens, the client reties, etc ... 

But TBH, I'm more curious of all possible situations where such "two different values for the same key" may occur in the first place in our workflows. Seem like "point's of failure" to me.

rogpeppe, line 66: The infinite loop scenario is a good point and something we definitely need to consider in our client code. 

The "two values for the same key" situation is not something that should ever happen in our normal flow,  but we have to consider the client as a potentially malicious attacker who is able to add caveats at will.  That's the situation we are considering here.

PR #17, opened 2014-12-12
bakery: depend on checkers rather than the other way around

rogpeppe:


rogpeppe:
Landing without review as this is an almost entirely mechanical change that has been approved in principle online.

PR #18, opened 2014-12-12
bakery: implement CheckAny

rogpeppe:
Factor CheckAny out from httpbakery.Check. There's no
new behaviour here.

Also (and trivial) rewrite []*macaroon.Macaroon to macaroon.Slice throughout
and change the argument order of httpbakery.Check.

### DIFF
--- a/bakery/service.go
+++ b/bakery/service.go
@@ -10,8 +10,9 @@ import (
 	"log"
 
 	"gopkg.in/errgo.v1"
-	"gopkg.in/macaroon-bakery.v0/bakery/checkers"
 	"gopkg.in/macaroon.v1"
+
+	"gopkg.in/macaroon-bakery.v0/bakery/checkers"
 )
 
 const debug = false
@@ -114,7 +115,7 @@ type Caveat struct {
 // If there is a verification error, it returns a VerificationError that
 // describes the error (other errors might be returned in other
 // circumstances).
-func (svc *Service) Check(ms []*macaroon.Macaroon, checker FirstPartyChecker) error {
+func (svc *Service) Check(ms macaroon.Slice, checker FirstPartyChecker) error {
 	if len(ms) == 0 {
 		return &VerificationError{
 			Reason: fmt.Errorf("no macaroons in slice"),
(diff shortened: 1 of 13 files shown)


tasdomas:
LGTM

httpbakery/client.go
```diff
@@ -354,7 +354,7 @@ func isVerificationError(err error) bool {
 // It adds all the standard caveat checkers to the given checker.
 //
 // It returns any attributes declared in the successfully validated request.
-func CheckRequest(svc *bakery.Service, req *http.Request, checker checkers.Checker, assert map[string]string) (map[string]string, error) {
+func CheckRequest(svc *bakery.Service, req *http.Request, assert map[string]string, checker checkers.Checker) (map[string]string, error) {
```
cmars, line 357: I'd consider creating a type for `map[string]string`. godocs and helper methods on that might make for a more fluent readable API.

bakery/service.go
```diff
@@ -141,6 +142,44 @@ func (svc *Service) Check(ms []*macaroon.Macaroon, checker FirstPartyChecker) er
 	return nil
 }
 
+// CheckAny checks that the given slice of slices contains at least
+// one macaroon minted by the given service, using checker to check
+// any first party caveats. It returns an error with a
+// *bakery.VerificationError cause if the macaroon verification failed.
+//
+// The assert map holds any required attributes of "declared" attributes,
```
cmars, line 150: A special type for the required declared attributes would be a nicer place to document this behavior & purpose.

httpbakery/client.go
rogpeppe, line 357: I'm not sure in this case. New types help when they actually have some associated semantics (and thus the possibility of useful methods), but they can also hinder, because they're incompatible with other named types with the same underlying type. In this case, this really is just a bundle of attributes and values, and the canonical way to declare that in Go is map[string] string. That means that the attributes are assignment-compatible with DeclaredChecker, for example, which they would not be if they were named.

I'd prefer to leave as is, at least for the time being.

alesstimec:
LGTM

PR #20, opened 2014-12-17
bakery: modify public keys in place

rogpeppe:
This allows a PublicKeyRing to be updated without growing indefinitely.

Also update the license to LGPL (with static linking exception) to conform
with company guidelines.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,26 +1,187 @@
-Copyright © 2014, Roger Peppe
-All rights reserved.
+Copyright © 2014, Roger Peppe, Canonical Inc.
 
-Redistribution and use in source and binary forms, with or without modification,
-are permitted provided that the following conditions are met:
+This software is licensed under the LGPLv3, included below.
 
-    * Redistributions of source code must retain the above copyright notice,
-      this list of conditions and the following disclaimer.
-    * Redistributions in binary form must reproduce the above copyright notice,
-      this list of conditions and the following disclaimer in the documentation
-      and/or other materials provided with the distribution.
-    * Neither the name of this project nor the names of its contributors
-      may be used to endorse or promote products derived from this software
-      without specific prior written permission.
+As a special exception to the GNU Lesser General Public License version 3
+("LGPL3"), the copyright holders of this Library give you permission to
+convey to a third party a Combined Work that links statically or dynamically
+to this Library without providing any Minimal Corresponding Source or
+Minimal Application Code as set out in 4d or providing the installation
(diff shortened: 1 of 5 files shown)


LICENSE
```diff
@@ -1,26 +1,187 @@
...
-LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
-A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
-OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
-SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED
-TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
-PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
-LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
-NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
-SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+Copyright © 2014, Roger Peppe, Canonical Inc.
```
mhilton, line 1: I thought we were Canonical Ltd.

bakery/keys.go
```diff
@@ -168,11 +168,19 @@ func (kr *PublicKeyRing) AddPublicKeyForLocation(loc string, prefix bool, key *P
...
 	kr.mu.Lock()
 	defer kr.mu.Unlock()
-	kr.publicKeys = append(kr.publicKeys, publicKeyRecord{
+	newr := publicKeyRecord{
 		url:    url,
 		prefix: prefix,
 		key:    *key,
-	})
+	}
+	for i := range kr.publicKeys {
```
mhilton, line 176: have you thought about using a map instead of a slice here?

mhilton:
:+1: LGTM with a couple of suggestions.

bakery/keys.go
rogpeppe, line 176: yes :)

since we have to iterate through all the keys for the lookup,
lookups are likely to be much more frequent than modifications
and iterating through a slice is faster than a map, i think
it's probably better as a slice. Better still as a prefix tree, but that's
too much for now.

PR #22, opened 2014-12-17
bakery/mgostorage: use canonical errgo import path

rogpeppe:


rogpeppe:
Merging as it's trivial.

PR #23, opened 2014-12-19
bakery: add support for need-declared third party caveat

rogpeppe:
This makes it possible to add a third-party caveat that requires
some declared caveats and be assured that (as long as they're
using the bakery package) they will be added.

### DIFF
--- a/bakery/service.go
+++ b/bakery/service.go
@@ -8,6 +8,7 @@ import (
 	"crypto/rand"
 	"fmt"
 	"log"
+	"strings"
 
 	"gopkg.in/errgo.v1"
 	"gopkg.in/macaroon.v1"
@@ -95,16 +96,6 @@ func (svc *Service) Location() string {
 // PublicKey returns the service's public key.
 func (svc *Service) PublicKey() *PublicKey {
 	return &svc.key.Public
-}
-
-// Caveat represents a condition that must be true for a check to
-// complete successfully. If Location is non-empty, the caveat must be
-// discharged by a third party at the given location.
-// This differs from macaroon.Caveat in that the condition
-// is not encrypted.
-type Caveat struct {
-	Location  string
-	Condition string
 }
 
 // Check checks that the given macaroons verify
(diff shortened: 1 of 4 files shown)


mhilton:
:+1: LGTM

bakery/service.go
```diff
@@ -260,13 +251,67 @@ func (svc *Service) Discharge(checker ThirdPartyChecker, id string) (*macaroon.M
...
+		if len(parts) != 2 {
+			return nil, errgo.Newf("declared caveat has no value")
+		}
+		declared[parts[0]] = true
+	}
+	// Add empty declarations for everything mentioned in need-declared
+	// that was not actually declared.
+	for _, d := range needDeclared {
+		if !declared[d] {
+			caveats = append(caveats, checkers.DeclaredCaveat(d, ""))
```
cmars, line 309: I assume this is a security measure to "seal" the macaroon from any further first-party declarations being added. Is that correct?

bakery/service_test.go
```diff
@@ -131,10 +143,93 @@ func (s *ServiceSuite) TestMacaroonPaperFig6FailsWithBindingOnTamperedSignature(
...
+	declared := checkers.InferDeclared(d)
+	c.Assert(declared, gc.DeepEquals, checkers.Declared{
+		"foo": "",
+		"bar": "",
+	})
+
+	// Make sure the macaroons actually check out correctly
+	// when provided with the declared checker.
+	err = firstParty.Check(d, checkers.New(declared))
+	c.Assert(err, gc.IsNil)
```
cmars, line 182: I feel like maybe this should be `NotNil`. If you're going to use empty-string as a catch-all for a missing declaration, maybe it should not pass the `needs-declared` check.

cmars:
Question and a comment. I found the empty-string values for undeclared caveats a little off-putting, and I'm not completely sure I understand the reason for them just from reading this diff.

bakery/service_test.go
```diff
@@ -131,10 +143,93 @@ func (s *ServiceSuite) TestMacaroonPaperFig6FailsWithBindingOnTamperedSignature(
 	}
 
 	// client makes request to ts.
-	err = ts.Check(d, strCompFirstPartyChecker(""))
+	err = ts.Check(d, strcmpChecker(""))
 	c.Assert(err, gc.ErrorMatches, "verification failed: signature mismatch after caveat verification")
 }
 
+func (s *ServiceSuite) TestNeedDeclared(c *gc.C) {
```
cmars, line 150: I'd like a test case with a macaroon slice that fails to verify because a `needs-declared` declaration is missing from the discharges. I see that you get empty-string defaults but it would be nice to detect the difference. Or am I missing something?

Wacky test cases I am curious about, just because it seems like they'd pass (and do we want them to?):
- Nested third-party caveats, which incrementally satisfy a `needs-declared`. I could imagine this being practical.
- Two third-party discharges that add `needs-declared` caveats, the discharges of which satisfy each other. Weird.
- A macaroon slice that adds a first-party caveat, with some caveats that satisfy a third-party `needs-declared` which it also adds.. the discharge for which does not duplicate any of the declarations. Weird and probably wrong but it would verify, wouldn't it?

Also, can you `needs-declared needs-declared needs-declared ...`?

[Edited]

bakery/service.go
rogpeppe, line 309: Yes. The important thing about need-declared is that it ensures that the required attributes are declared.

bakery/service_test.go
rogpeppe, line 182: As discussed over breakfast, I think it's reasonable to be able to add a need-declared attribute without necessarily requiring the caveat discharger to explicitly add it. This means that we can add a need-declared attribute that can work against an old version of a server without necessarily breaking the discharging. The worst that can happen is that the attribute ends up declared as an empty string, which seems like a reasonable default value to me.

bakery/service_test.go
rogpeppe, line 150: The nested third-party caveat need-declared thing is an interesting possibility, but one that can be added later, I think. I've written another test to check with two third party discharges. I think the last example is sufficiently tested by the existing lower level checks in checkers.

With nested need-declared, the inner one won't be recognised - it's recognised directly by the bakery third party checker and nowhere else.

bakery/service.go
rogpeppe, line 309: Yes - this is the raison d'etre of this caveat.

cmars:
LGTM. Thanks for clarifying!

PR #24, opened 2015-01-06
httpbakery: remove DefaultHTTPClient variable

rogpeppe:
It is too easy to abuse, and it's easy to create an appropriate client
when needed. We replace it with the NewHTTPClient function
which does not have the same issue with mutable global state.


mhilton:
:+1: LGTM

frankban:
:+1: thanks

PR #25, opened 2015-01-21
httpbakery: add DoWithBody function

rogpeppe:
We need this so that we can retry the HTTP request without
losing the body.

### DIFF
--- a/httpbakery/client.go
+++ b/httpbakery/client.go
@@ -64,6 +64,12 @@ type WaitResponse struct {
 // any required discharge macaroons will be acquired,
 // and the request will be repeated with those attached.
 //
+// Note that because the request may be retried, no
+// body may be provided in the http request (otherwise
+// the contents will be lost when retrying). For requests
+// with a body (for example PUT or POST methods),
+// use DoWithBody instead.
+//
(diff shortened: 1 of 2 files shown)


httpbakery/client.go
```diff
@@ -71,6 +77,20 @@ type WaitResponse struct {
 // function is called with a URL to be opened in a
 // web browser.
 func Do(client *http.Client, req *http.Request, visitWebPage func(url *url.URL) error) (*http.Response, error) {
+	if req.Body != nil {
+		return nil, fmt.Errorf("body unexpectedly provided in request - use DoWithBody")
+	}
+	return DoWithBody(client, req, noBody, visitWebPage)
+}
+
+func noBody() (io.ReadCloser, error) {
```
alesstimec, line 86: noBody is used only once and could be defined inline..

alesstimec:
LGTM! thank you.

cmars:
LGTM, thanks!

httpbakery/client.go
rogpeppe, line 86: I'll leave it as is if that's OK. Then it's obvious that it doesn't allocate.

PR #26, opened 2015-01-21
Simple keygen utility for bakery

mhilton:
Simple utility to generate  bakery keys.

### DIFF
--- a/cmd/bakery-keygen/main.go
+++ b/cmd/bakery-keygen/main.go
@@ -2,33 +2,21 @@ package main
 
 import (
 	"encoding/json"
-	"flag"
 	"fmt"
 	"os"
 
 	"gopkg.in/macaroon-bakery.v0/bakery"
 )
 
-var format = flag.String("f", "json", "output format (json, go)")
-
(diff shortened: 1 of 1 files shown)


cmd/bakery-keygen/main.go
```diff
@@ -0,0 +1,34 @@
...
+var format = flag.String("f", "json", "output format (json, go)")
+
+func main() {
+	flag.Parse()
+	kp, err := bakery.GenerateKey()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "cannot generate key: %s\n", err)
+		os.Exit(1)
+	}
+	switch *format {
```
rogpeppe: I'd suggest we keep things ultra simple here and just use JSON.
No particular need to make the output format configurable, I think.
Then we don't need to parse arguments at all.

rogpeppe:
LGTM thanks, modulo one suggestion above.

rogpeppe:
Still LGTM, thanks.

jrwren:
:+1:

PR #27, opened 2015-01-23
all: use loggo rather than log for logging

rogpeppe:
Also update the dependencies.tsv file, which hasn't been updated for a while.

### DIFF
--- a/bakery/storage.go
+++ b/bakery/storage.go
@@ -40,7 +40,7 @@ type memStorage struct {
 }
 
 func (s *memStorage) Put(location, item string) error {
-	logf("storage.Put[%q] %q", location, item)
+	logger.Infof("storage.Put[%q] %q", location, item)
 	s.mu.Lock()
 	defer s.mu.Unlock()
 	s.values[location] = item
@@ -52,10 +52,10 @@ func (s *memStorage) Get(location string) (string, error) {
 	defer s.mu.Unlock()
 	item, ok := s.values[location]
 	if !ok {
-		logf("storage.Get[%q] -> not found", location)
+		logger.Infof("storage.Get[%q] -> not found", location)
 		return "", ErrNotFound
 	}
-	logf("storage.Get[%q] -> %q", location, item)
+	logger.Infof("storage.Get[%q] -> %q", location, item)
 	return item, nil
 }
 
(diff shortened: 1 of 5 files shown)


jrwren:
:+1: :man:

bakery/storage.go
```diff
@@ -52,10 +52,10 @@ func (s *memStorage) Get(location string) (string, error) {
 	defer s.mu.Unlock()
 	item, ok := s.values[location]
 	if !ok {
-		logf("storage.Get[%q] -> not found", location)
+		logger.Infof("storage.Get[%q] -> not found", location)
```
mhilton, line 55: Should this possibly be higher than Info Error maybe?

httpbakery/client.go
```diff
@@ -147,9 +149,9 @@ func relativeURL(base, new string) (*url.URL, error) {
 }
 
 func (ctxt *clientContext) do(req *http.Request, getBody BodyGetter) (*http.Response, error) {
-	log.Printf("client do %s %s {", req.Method, req.URL)
+	logger.Debugf("client do %s %s {", req.Method, req.URL)
 	resp, err := ctxt.do1(req, getBody)
-	log.Printf("} -> error %#v", err)
+	logger.Infof("} -> error %#v", err)
```
mhilton: Might this look a little odd if logging at info level or above, the curly brackets won't match.

httpbakery/client.go
```diff
@@ -271,7 +273,7 @@ func (ctxt *clientContext) obtainThirdPartyDischarge(originalLocation string, ca
 	if err == nil {
 		return resp.Macaroon, nil
 	}
-	log.Printf("discharge post got error %#v", err)
+	logger.Debugf("discharge post got error %#v", err)
```
mhilton: Is this really a debug level message?

mhilton:
:+1: LGTM, but check you're happy with the log levels.

httpbakery/client.go
rogpeppe: Ha, they should both be at debug level.

httpbakery/client.go
rogpeppe: We should probably just delete the log message - the error will be returned and should be logged anyway by our caller.

bakery/storage.go
rogpeppe, line 55: Not sure. The error should be logged elsewhere anyway. It seems to me like the same level as the other storage access log messages is appropriate.

PR #28, opened 2015-01-25
httpbakery: add PublicKeyForLocation

mattyw:
Not totally sure it belongs in this repo: But: Easy to ask forgiveness than permission etc

The idea is:

A server using the DischargeHandler serves up the public key on a particular path and encoded a certain way. It seems at least partly sensible for this repo to provide a way of getting that public key in a usable format

### DIFF
--- a/bakerytest/bakerytest.go
+++ b/bakerytest/bakerytest.go
@@ -62,6 +62,7 @@ func (d *Discharger) Close() {
 
 // Location returns the location of the discharger, suitable
 // for setting as the location in a third party caveat.
+// This will be the URL of the server.
 func (d *Discharger) Location() string {
 	return d.Service.Location()
 }
--- a/httpbakery/client.go
+++ b/httpbakery/client.go
@@ -127,6 +127,33 @@ func DischargeAll(m *macaroon.Macaroon, client *http.Client, visitWebPage func(u
 	return bakery.DischargeAll(m, ctxt.obtainThirdPartyDischarge)
 }
 
+// PublicKeyForLocation returns the public key from a macaroon
+// discharge server running at the given location URL.
+// Note that this is insecure if an http: URL scheme is used.
+func PublicKeyForLocation(client *http.Client, url string) (*bakery.PublicKey, error) {
+	url = url + "/publickey"
+	resp, err := client.Get(url)
+	if err != nil {
+		return nil, errgo.Notef(err, "cannot get public key from %q", url)
+	}
+	if resp.StatusCode != http.StatusOK {
+		return nil, errgo.Newf("cannot get public key from %q: got status %s", url, resp.Status)
+	}
+	defer resp.Body.Close()
+	data, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, errgo.Notef(err, "failed to read response body from %q", url)
+	}
+	var pubkey struct {
+		PublicKey *bakery.PublicKey
+	}
+	err = json.Unmarshal(data, &pubkey)
+	if err != nil {
+		return nil, errgo.Notef(err, "failed to decode response from %q", url)
+	}
+	return pubkey.PublicKey, nil
+}
+
 type clientContext struct {
 	client       *http.Client
 	visitWebPage func(*url.URL) error
--- a/httpbakery/client_test.go
+++ b/httpbakery/client_test.go
@@ -156,6 +156,45 @@ func (s *ClientSuite) TestMacaroonCookiePath(c *gc.C) {
 	assertCookieCount("/foo/bar/baz", 1)
 }
 
+func (s *ClientSuite) TestPublicKey(c *gc.C) {
+	d := bakerytest.NewDischarger(nil, noCaveatChecker)
+	defer d.Close()
+	client := httpbakery.NewHTTPClient()
+	publicKey, err := httpbakery.PublicKeyForLocation(client, d.Location())
+	c.Assert(err, gc.IsNil)
+	expectedKey := d.Service.PublicKey()
+	c.Assert(publicKey, gc.DeepEquals, expectedKey)
+}
+
+func (s *ClientSuite) TestPublicKeyWrongURL(c *gc.C) {
+	client := httpbakery.NewHTTPClient()
+	_, err := httpbakery.PublicKeyForLocation(client, "http://localhost:0")
+	c.Assert(err, gc.ErrorMatches,
+		`cannot get public key from "http://localhost:0/publickey": Get http://localhost:0/publickey: dial tcp 127.0.0.1:0: connection refused`)
+}
+
+func (s *ClientSuite) TestPublicKeyReturnsInvalidJson(c *gc.C) {
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		fmt.Fprintln(w, "BADJSON")
+	}))
+	defer ts.Close()
+	client := httpbakery.NewHTTPClient()
+	_, err := httpbakery.PublicKeyForLocation(client, ts.URL)
+	c.Assert(err, gc.ErrorMatches,
+		fmt.Sprintf(`failed to decode response from "%s/publickey": invalid character 'B' looking for beginning of value`, ts.URL))
+}
+
+func (s *ClientSuite) TestPublicKeyReturnsStatusInternalServerError(c *gc.C) {
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusInternalServerError)
+	}))
+	defer ts.Close()
+	client := httpbakery.NewHTTPClient()
+	_, err := httpbakery.PublicKeyForLocation(client, ts.URL)
+	c.Assert(err, gc.ErrorMatches,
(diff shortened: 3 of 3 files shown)


bakerytest/bakerytest.go
```diff
@@ -73,3 +73,8 @@ func (d *Discharger) PublicKeyForLocation(loc string) (*bakery.PublicKey, error)
 	}
 	return nil, bakery.ErrNotFound
 }
+
+// URL returns the URL the discharger is using.
```
rogpeppe: I don't think this function is necessary - we can use d.Service.Location() instead (perhaps we should document that on the Discharger type).

httpbakery/client.go
```diff
@@ -125,6 +125,28 @@ func DischargeAll(m *macaroon.Macaroon, client *http.Client, visitWebPage func(u
 	return bakery.DischargeAll(m, ctxt.obtainThirdPartyDischarge)
 }
 
+// GetPublicKey returns the public key provided by a server which is using
+// httpbakery.AddDischargeHandler.
+func GetPublicKey(url string) (*bakery.PublicKey, error) {
```
rogpeppe: // PublicKeyForLocation returns the public key from a macaroon
// discharge server running at the given location URL.
// Note that this is insecure if an http: URL scheme is used.

?

I think we should probably pass the http client in here too,
so that we're free to use httpbakery.Do in the future if
we want.

httpbakery/client.go
```diff
@@ -125,6 +125,28 @@ func DischargeAll(m *macaroon.Macaroon, client *http.Client, visitWebPage func(u
...
+	resp, err := http.Get(url + "/publickey")
+	if err != nil {
+		return nil, errgo.Notef(err, "failed to make request for public key")
+	}
+	if resp.StatusCode != http.StatusOK {
+		return nil, errgo.Newf("server returned error code %v", resp.StatusCode)
+	}
+	defer resp.Body.Close()
+	decoder := json.NewDecoder(resp.Body)
```
rogpeppe: In general for JSON response bodies, I prefer to do an ioutil.ReadAll - that way we don't need to abandon the network connection if the JSON happens to have a trailing newline.

httpbakery/client_test.go
```diff
@@ -101,6 +101,16 @@ func (s *ClientSuite) TestRepeatedRequestWithBody(c *gc.C) {
...
 }
 
+func (s *ClientSuite) TestGetPublicKey(c *gc.C) {
+	d := bakerytest.NewDischarger(nil, noCaveatChecker)
+	defer d.Close()
+
+	publicKey, err := httpbakery.GetPublicKey(d.URL())
+	c.Assert(err, gc.IsNil)
+	expectedKey, err := d.PublicKeyForLocation(d.Location())
+	c.Assert(publicKey, gc.DeepEquals, expectedKey)
```
rogpeppe, line 166: c.Assert(err, gc.IsNil)

httpbakery/client_test.go
```diff
@@ -101,6 +101,16 @@ func (s *ClientSuite) TestRepeatedRequestWithBody(c *gc.C) {
 	c.Assert(bodyReader.byteCount, gc.Equals, len(bodyText)*2)
 }
 
+func (s *ClientSuite) TestGetPublicKey(c *gc.C) {
+	d := bakerytest.NewDischarger(nil, noCaveatChecker)
+	defer d.Close()
+
+	publicKey, err := httpbakery.GetPublicKey(d.URL())
+	c.Assert(err, gc.IsNil)
+	expectedKey, err := d.PublicKeyForLocation(d.Location())
```
rogpeppe: expectedKey := d.Service.PublicKey()

rogpeppe:
This totally belongs in this repo - thanks for the PR!

I'd rename the PR to follow the usual "package: description"
convention. e.g. httpbakery: add PublicKeyForLocation

Looks great in general, modulo a few suggestions above.

httpbakery/client.go
```diff
@@ -127,6 +127,32 @@ func DischargeAll(m *macaroon.Macaroon, client *http.Client, visitWebPage func(u
 	return bakery.DischargeAll(m, ctxt.obtainThirdPartyDischarge)
 }
 
+// PublicKeyForLocation returns the public key from a macaroon
+// discharge server running at the given location URL.
+// Note that this is insecure if an http: URL scheme is used.
+func GetPublicKeyForLocation(client *http.Client, url string) (*bakery.PublicKey, error) {
```
rogpeppe: s/Get// ?

(especially as that's what the comment says)

httpbakery/client.go
```diff
@@ -127,6 +127,32 @@ func DischargeAll(m *macaroon.Macaroon, client *http.Client, visitWebPage func(u
...
+	resp, err := client.Get(url + "/publickey")
+	if err != nil {
+		return nil, errgo.Notef(err, "failed to make request for public key")
+	}
+	if resp.StatusCode != http.StatusOK {
+		return nil, errgo.Newf("server returned error code %v", resp.StatusCode)
```
rogpeppe: errgo.Newf("cannot get public key from %q: got status %s", url, resp.Status)

?
(you'll probably want to do "url += "/publickey" before this (before client.Get in fact) so that the url looks right in the message)

httpbakery/client_test.go
```diff
@@ -156,6 +156,16 @@ func (s *ClientSuite) TestMacaroonCookiePath(c *gc.C) {
 	assertCookieCount("/foo/bar/baz", 1)
 }
 
+func (s *ClientSuite) TestGetPublicKey(c *gc.C) {
```
rogpeppe: Perhaps some tests for the error conditions too?

rogpeppe:
Nearly there, thanks! A few more comments above.

httpbakery/client.go
```diff
@@ -127,6 +127,33 @@ func DischargeAll(m *macaroon.Macaroon, client *http.Client, visitWebPage func(u
...
 }
 
+// PublicKeyForLocation returns the public key from a macaroon
+// discharge server running at the given location URL.
+// Note that this is insecure if an http: URL scheme is used.
+func PublicKeyForLocation(client *http.Client, url string) (*bakery.PublicKey, error) {
+	url = url + "/publickey"
+	resp, err := client.Get(url)
+	if err != nil {
+		return nil, errgo.Newf("cannot get public key from %q: error %s", url, err)
```
rogpeppe: s/error //

httpbakery/client_test.go
```diff
@@ -156,6 +156,45 @@ func (s *ClientSuite) TestMacaroonCookiePath(c *gc.C) {
...
+	c.Assert(err, gc.IsNil)
+	expectedKey := d.Service.PublicKey()
+	c.Assert(publicKey, gc.DeepEquals, expectedKey)
+}
+
+func (s *ClientSuite) TestPublicKeyWrongURL(c *gc.C) {
+	client := httpbakery.NewHTTPClient()
+	_, err := httpbakery.PublicKeyForLocation(client, "http://localhost:0")
+	c.Assert(err, gc.ErrorMatches,
+		`cannot get public key from "http://localhost:0/publickey": Get http://localhost:0/publickey: dial tcp 127.0.0.1:0: connection refused`)
```
rogpeppe, line 173: OK, so it seems that the url in our annotation is redundant. Let's leave it out.

httpbakery/client_test.go
```diff
@@ -156,6 +156,45 @@ func (s *ClientSuite) TestMacaroonCookiePath(c *gc.C) {
...
+
+func (s *ClientSuite) TestPublicKeyReturnsInvalidJson(c *gc.C) {
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		fmt.Fprintln(w, "BADJSON")
+	}))
+	defer ts.Close()
+	client := httpbakery.NewHTTPClient()
+	_, err := httpbakery.PublicKeyForLocation(client, ts.URL)
+	c.Assert(err, gc.ErrorMatches,
+		`failed to decode response from server: invalid character 'B' looking for beginning of value`)
```
rogpeppe: ... on  the other hand, let's put the URL into this one.

rogpeppe: (and the "failed to read response body" one)

rogpeppe:
LGTM modulo three more trivials above. Thanks!

httpbakery/client.go
```diff
@@ -127,6 +127,33 @@ func DischargeAll(m *macaroon.Macaroon, client *http.Client, visitWebPage func(u
...
+// discharge server running at the given location URL.
+// Note that this is insecure if an http: URL scheme is used.
+func PublicKeyForLocation(client *http.Client, url string) (*bakery.PublicKey, error) {
+	url = url + "/publickey"
+	resp, err := client.Get(url)
+	if err != nil {
+		return nil, errgo.Notef(err, "cannot get public key from %q", url)
+	}
+	if resp.StatusCode != http.StatusOK {
+		return nil, errgo.Newf("cannot get public key from %q: got status: %d", url, resp.StatusCode)
```
rogpeppe: return nil, errgo.Newf("cannot get public key from %q: got status %s", url, resp.Status)

better to print the whole status, i think (and the : was unnecessary)

rogpeppe:
LGTM again, with one trivial

PR #31, opened 2015-02-12
example: fix for change to WriteDischargeRequiredError

rogpeppe:


rogpeppe:
trivial, so merging immediately.

PR #32, opened 2015-02-19
httpbakery: test macaraq on discharge

rogpeppe:
I wanted to verify that this functionality actually works as expected,
where a third party discharger issues a discharge but only after
issuing a macaroon with a third party caveat itself.

### DIFF
--- a/httpbakery/client_test.go
+++ b/httpbakery/client_test.go
@@ -9,7 +9,8 @@ import (
 	"net/url"
 	"strings"
 
-	jujuTesting "github.com/juju/testing"
+	jujutesting "github.com/juju/testing"
+	"github.com/juju/utils/jsonhttp"
 	gc "gopkg.in/check.v1"
 	"gopkg.in/macaroon.v1"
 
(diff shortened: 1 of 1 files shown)


httpbakery/client_test.go
```diff
@@ -219,21 +275,17 @@ func (r *readCounter) Read(buf []byte) (int, error) {
...
 }
 
-func newService(c *gc.C, location string, locator bakery.PublicKeyLocator) *bakery.Service {
+func newService(location string, locator bakery.PublicKeyLocator) *bakery.Service {
 	svc, err := bakery.NewService(bakery.NewServiceParams{
 		Location: location,
 		Locator:  locator,
 	})
-	c.Assert(err, gc.IsNil)
+	if err != nil {
```
jrwren, line 283: Not a blocker: I would like to know any guidelines on using panic v. passing gc.C in a test builder function like this.

jrwren:
:+1:

httpbakery/client_test.go
rogpeppe, line 283: I tend to use panic the logic is extremely unlikely to fail and if it does that's nothing to
do with what's being tested.

In this case, NewService can fail only because the system random number
generator returns an error, which will never happen on any of the usual
platforms AFAIK.

mhilton:
:+1: LGTM

PR #33, opened 2015-02-20
bakery: add support for self-discharged 3rd party caveats

rogpeppe:
This also includes a revamp of the httpbakery client interface
to make it based around a Client type, so it's easy to add
the Key field without requiring a new argument to be added
everywhere.

As a drive-by change, we also remove the redundant /create
endpoint in the discharger.


rogpeppe:
Closing to retarget at macaroon-bakery.v1

PR #38, opened 2015-04-07
httpbakery: implement OpenWebBrowser

rogpeppe:
We don't add any tests because the tests are going to be very awkward
and system dependent, and can't test the thing we post care about,
which is that a web browser gets started.

To QA, compile the following program and run it on whatever platform
you are on. It should open up the macaroon-bakery documentation.
For bonus points, play with your $PATH to check the error conditions.

```
package main
import (
    "log"
    "net/url"
    "gopkg.in/macaroon-bakery.v0/httpbakery"
)

func main() {
    u, err := url.Parse("http://godoc.org/gopkg.in/macaroon-bakery.v0")
    if err != nil {
        log.Fatal(err)
    }
    err = httpbakery.OpenWebBrowser(u)
    if err != nil {
        log.Fatal(err)
    }
    log.Printf("success")
}
```

We also update the dependencies and remove the outdated dependency on
code.google.com/p/go.net .


alesstimec:
LGTM

frankban:
:+1:

PR #39, opened 2015-04-07
httpbakery: fix OpenWebBrowser for Windows

rogpeppe:
To QA, compile the following program and run it on whatever platform
you are on. It should open up the macaroon-bakery documentation.
For bonus points, play with your $PATH to check the error conditions.

```
package main
import (
    "log"
    "net/url"
    "gopkg.in/macaroon-bakery.v0/httpbakery"
)

func main() {
    u, err := url.Parse("http://godoc.org/gopkg.in/macaroon-bakery.v0")
    if err != nil {
        log.Fatal(err)
    }
    err = httpbakery.OpenWebBrowser(u)
    if err != nil {
        log.Fatal(err)
    }
    log.Printf("success")
}
```

### DIFF
--- a/httpbakery/browser.go
+++ b/httpbakery/browser.go
@@ -1,0 +1,68 @@
+package httpbakery
+
+import (
+	"bytes"
+	"fmt"
+	"net/url"
+	"os"
+	"os/exec"
+	"runtime"
+	"strings"
+
+	"gopkg.in/errgo.v1"
+)
+
+var browser = map[string]string{
+	"linux":   "sensible-browser",
+	"darwin":  "open",
+	"freebsd": "xdg-open",
+	"netbsd":  "xdg-open",
+	"openbsd": "xdg-open",
+}
+
+// OpenWebBrowser opens a web browser at the
+// given URL. If the OS is not recognised, the URL
+// is just printed to standard output.
+func OpenWebBrowser(url *url.URL) error {
+	var args []string
+	if runtime.GOOS == "windows" {
+		// Windows is special because the start command is
+		// built into cmd.exe and hence requires the argument
+		// to be quoted.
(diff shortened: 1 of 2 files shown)


httpbakery/browser.go
```diff
@@ -0,0 +1,68 @@
...
+	"netbsd":  "xdg-open",
+	"openbsd": "xdg-open",
+}
+
+// OpenWebBrowser opens a web browser at the
+// given URL. If the OS is not recognised, the URL
+// is just printed to standard output.
+func OpenWebBrowser(url *url.URL) error {
+	var args []string
+	if true || runtime.GOOS == "windows" {
```
jrwren: if true || ?

httpbakery/browser.go
```diff
@@ -0,0 +1,68 @@
...
+			return nil
+		}
+		if err != exec.ErrNotFound {
+			if _, ok := err.(*exec.ExitError); ok {
+				return errgo.Newf("cannot open web browser: %s", bytes.TrimSpace(data))
+			}
+			return errgo.Notef(err, "cannot open web browser")
+		}
+	}
+	fmt.Fprintf(os.Stderr, "Please visit this web page:\n%s\n", url)
```
jrwren, line 50: This output will not display in the above error cases. Should it display in the case that the web browser cannot open?

jrwren:
:+1:  with a couple of questions.

httpbakery/browser.go
rogpeppe: oops :)

httpbakery/browser.go
rogpeppe, line 50: It's a good question. If you've got a dud web browser, should we just ignore that fact and just print the web address? Currently we do that only if the web browser isn't found. Up for debate here.

frankban:
:+1:

PR #40, opened 2015-04-08
bakerytest: pass http.Request to discharger caveat checker

rogpeppe:
This allows us to write checkers that depend on the client request,
such as the cookies that are present.


mhilton:
:+1: LGTM

jrwren:
:+1:

PR #41, opened 2015-04-10
httpbakery: always print URL for browser and do not wait

rogpeppe:
The web browser commands are uncertain. They may block forever,
or fail for arbitrary reasons; so we always print the URL and we do
not block waiting for the command to exit.


mattyw:
LGTM - But I think it might be useful to provide this as an exported function at a later date

urosj:
:+1:

PR #42, opened 2015-04-13
httpbakery: new Client type

rogpeppe:
This means that both http.Client and httpbakery.Client implement
the same interface. It also makes it easier to add client-specific
information in the future.

### DIFF
--- a/httpbakery/client.go
+++ b/httpbakery/client.go
@@ -56,6 +56,13 @@ func IsInteractionError(err error) bool {
 	return ok
 }
 
+// WaitResponse holds the type that should be returned
+// by an HTTP response made to a WaitURL
+// (See the ErrorInfo type).
+type WaitResponse struct {
+	Macaroon *macaroon.Macaroon
+}
+
 // NewHTTPClient returns an http.Client that ensures
 // that headers are sent to the server even when the
 // server redirects a GET request. The returned client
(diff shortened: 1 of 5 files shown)


httpbakery/client.go
```diff
@@ -88,11 +95,27 @@ func NewHTTPClient() *http.Client {
...
 }
 
-// WaitResponse holds the type that should be returned
-// by an HTTP response made to a WaitURL
-// (See the ErrorInfo type).
-type WaitResponse struct {
-	Macaroon *macaroon.Macaroon
+// Client holds the context for making HTTP requests
+// that automatically acquire and discharge macaroons.
+type Client struct {
```
mhilton, line 100: It'd be  really good if the zero value just  worked, perhaps have a httpbakery.DefaultHTTPClient that it can use?

httpbakery/client.go
```diff
@@ -88,11 +95,27 @@ func NewHTTPClient() *http.Client {
...
+
+	// VisitWebPage is called when the authorization process
+	// requires user interaction, and should cause the
+	// given URL to be opened in a web browser.
+	// If this is nil, no interaction will be allowed.
+	VisitWebPage func(*url.URL) error
+}
+
+// NewClient returns a new Client containing an HTTP client
+// created with NewHTTPClient leaves all other fields zero.
```
frankban: s/leaves/leaving?

httpbakery/client.go
```diff
@@ -403,12 +406,15 @@ func (ctxt *clientContext) interact(location, visitURLStr, waitURLStr string) (*
 	if err != nil {
 		return nil, errgo.Notef(err, "cannot make relative wait URL")
 	}
-	if err := ctxt.visitWebPage(visitURL); err != nil {
+	if c.VisitWebPage == nil {
+		return nil, errgo.New("interaction required but not possible")
```
frankban: InteractionError?

frankban:
:+1: ty!

httpbakery/client.go

httpbakery/client.go
rogpeppe, line 100: As discussed, this would be awkward because we'd either need to copy the Client struct to add the default client or pass the http client around in parameters from method to method. Better as is, I think.

mhilton:
:+1: LGTM

httpbakery/client.go

PR #43, opened 2015-04-16
Re-add httpbakery.BodyGetter and httpbakery.SeekerBody

mhilton:
These symbols have been removed, but are required in the exported
interface of other libraries. Compatible versions have been added
with deprecated  warnings.


rogpeppe:
Closing as this wasn't necessary in the end.

PR #44, opened 2015-04-17
revert last commit (httpbakery: new Client type)

rogpeppe:
The client API change caused all kinds of problems with
clients of the bakery. Instead of breaking the API,
we keep it the same but create a new (stable) version
of the bakery at gopkg.in/macaroon-bakery.v1
that incorporates the changes that are reverted here.


mhilton:
:+1: Thank You!

PR #48, opened 2015-05-15
bakery: add support for self-discharged 3rd party caveats

rogpeppe:
This will support agent non-interactive authorization. The changes are intended to be backwardly compatible - please sing out if any API-breaking changes are seen.

### DIFF
--- a/bakery/codec.go
+++ b/bakery/codec.go
@@ -8,8 +8,6 @@ import (
 	"fmt"
 
 	"golang.org/x/crypto/nacl/box"
-
-	"gopkg.in/macaroon-bakery.v1/bakery/checkers"
 )
 
 type caveatIdRecord struct {
@@ -28,28 +26,18 @@ type caveatId struct {
 // boxEncoder encodes caveat ids confidentially to a third-party service using
 // authenticated public key encryption compatible with NaCl box.
 type boxEncoder struct {
-	locator PublicKeyLocator
-	key     *KeyPair
+	key *KeyPair
 }
 
-// newBoxEncoder creates a new boxEncoder with the given public key pair and
-// third-party public key locator function.
-func newBoxEncoder(locator PublicKeyLocator, key *KeyPair) *boxEncoder {
+// newBoxEncoder creates a new boxEncoder that uses the given public key pair.
+func newBoxEncoder(key *KeyPair) *boxEncoder {
 	return &boxEncoder{
-		key:     key,
-		locator: locator,
+		key: key,
 	}
 }
 
-func (enc *boxEncoder) encodeCaveatId(cav checkers.Caveat, rootKey []byte) (string, error) {
-	if cav.Location == "" {
-		return "", fmt.Errorf("cannot make caveat id for first party caveat")
-	}
-	thirdPartyPub, err := enc.locator.PublicKeyForLocation(cav.Location)
-	if err != nil {
-		return "", err
-	}
-	id, err := enc.newCaveatId(cav, rootKey, thirdPartyPub)
+func (enc *boxEncoder) encodeCaveatId(condition string, rootKey []byte, thirdPartyPub *PublicKey) (string, error) {
+	id, err := enc.newCaveatId(condition, rootKey, thirdPartyPub)
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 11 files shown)


bakery/discharge.go
```diff
@@ -5,18 +5,36 @@ import (
...
 	m *macaroon.Macaroon,
 	getDischarge func(firstPartyLocation string, cav macaroon.Caveat) (*macaroon.Macaroon, error),
 ) (macaroon.Slice, error) {
+	return DischargeAllWithKey(m, getDischarge, nil)
+}
+
+// DischargeAllWithKey is like DischargeAll except that the selfKey
+// parameter may optionally hold the key of the client, in which case it
+// will be used to discharge any third party caveats with the special
+// location "self". In this case, the caveat itself must be "true". This
```
mhilton: I still find "self" confusing as a name. Whan you add the third party caveat to a macaroon you'll be addressing the macaroon to self, but its not yourself that it is going to but the client.

httpbakery/client.go
```diff
@@ -98,16 +98,23 @@ func NewHTTPClient() *http.Client {
...
+	*http.Client
 
 	// VisitWebPage is called when the authorization process
 	// requires user interaction, and should cause the
 	// given URL to be opened in a web browser.
 	// If this is nil, no interaction will be allowed.
 	VisitWebPage func(*url.URL) error
+
+	// Key holds the client's key.
+	// If set, the client will be try to discharge third party
```
mhilton: Edit .+-x/be/d

mhilton:
:+1: LGTM with a couple of comments.

bakery/discharge.go
rogpeppe: This is true. I was coming from the p.o.v. that caveat locations are interpreted by the client, so "self" seems to work OK there. Better suggestions? "client" perhaps?

mhilton: "client" was going to be my suggestion. I can't think of anything really good, but I think client makes it clear when minting and discharging.

bakery/codec.go
```diff
@@ -28,28 +26,19 @@ type caveatId struct {
...
 // authenticated public key encryption compatible with NaCl box.
 type boxEncoder struct {
-	locator PublicKeyLocator
-	key     *KeyPair
+	key *KeyPair
 }
 
 // newBoxEncoder creates a new boxEncoder with the given public key pair and
 // third-party public key locator function.
-func newBoxEncoder(locator PublicKeyLocator, key *KeyPair) *boxEncoder {
```
urosj, line 37: Description needs to change, there's no locator anymore.

rogpeppe, line 37: good catch.

bakery/discharge.go
urosj: Client sounds to general as we're not targeting "any" client. But I'm not too keen on "sourceClient" either, as it might not be the source of the macaroon. No proper suggestions currently :(

bakery/service.go
```diff
@@ -257,9 +292,30 @@ func (svc *Service) Discharge(checker ThirdPartyChecker, id string) (*macaroon.M
...
+	m, err := macaroon.New(rootKey, id, "")
+	if err != nil {
+		return nil, nil, errgo.Mask(err)
+	}
+	return m, caveats, nil
+}
+
+// Discharge calls Discharge with the service's key and uses the service
+// to add any returned caveats to the discharge macaroon.
+func (svc *Service) Discharge(checker ThirdPartyChecker, id string) (*macaroon.Macaroon, error) {
```
cmars, line 309: This changes the API, doesn't it? Should we call this v2?

rogpeppe, line 309: I think this method still has exactly the same signature as it did before, doesn't it?
We're allowed to add functions and still keep the major version the same.

cmars, line 309: Didn't see the method pointer.. sorry, forget it.

bakery/discharge.go
rogpeppe: Updated to use "local" instead. Let's see how that feels.

httpbakery/client.go

bakery/codec.go
rogpeppe, line 37: done

bakery/codec.go
```diff
@@ -28,28 +26,18 @@ type caveatId struct {
 // boxEncoder encodes caveat ids confidentially to a third-party service using
 // authenticated public key encryption compatible with NaCl box.
 type boxEncoder struct {
-	locator PublicKeyLocator
```
cmars, line 31: I'm very happy with this moving to `Service`.

httpbakery/discharge.go
```diff
@@ -71,7 +60,6 @@ func AddDischargeHandler(mux *http.ServeMux, rootPath string, svc *bakery.Servic
 		checker: checker,
 	}
 	mux.Handle(path.Join(rootPath, "discharge"), handleJSON(d.serveDischarge))
-	mux.Handle(path.Join(rootPath, "create"), handleJSON(d.serveCreate))
```
cmars, line 74: This was a little concerning to me, but perfectly fine if you've been using the (http)bakery all along on client and server-side. I suppose someone could handle web API compatibility in ~~your~~ their services if they really depended on this (which seems a remote possibility). Good move I think.

cmars:
LGTM. Looking forward to this, thanks!

PR #49, opened 2015-05-20
bakery: add test to make sure discharge macaroons cannot be used for normal authorization

rogpeppe:
A conversation on IRC left me with a lingering doubt as to whether this
might be a problem. The test reassures me that it's not.

### DIFF
--- a/bakery/service.go
+++ b/bakery/service.go
@@ -296,7 +296,9 @@ func Discharge(key *KeyPair, checker ThirdPartyChecker, id string) (*macaroon.Ma
 		return nil, nil, errgo.Mask(err, errgo.Any)
 	}
 	// Note that the discharge macaroon does not need to
-	// be stored persistently.
+	// be stored persistently. Indeed, it would be a problem if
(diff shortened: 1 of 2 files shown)


mhilton:
LGTM

bakery/service.go
```diff
@@ -296,7 +296,9 @@ func Discharge(key *KeyPair, checker ThirdPartyChecker, id string) (*macaroon.Ma
 		return nil, nil, errgo.Mask(err, errgo.Any)
 	}
 	// Note that the discharge macaroon does not need to
-	// be stored persistently.
+	// be stored persistently. Indeed, it would be a problem if
```
mattyw, line 299: This comment makes me uneasy. If a bug was introduced that stored them persistently are we saying it would cause security problems?

mattyw:
LGTM

bakery/service.go
rogpeppe, line 299: As discussed online, this test is here in part to protect against that bug possibility.

PR #52, opened 2015-06-22
bakery/mgostorage: do not copy sessions

rogpeppe:
This means we can create a new mgostorage without incurring
a database round trip.

### DIFF
--- a/bakery/mgostorage/storage.go
+++ b/bakery/mgostorage/storage.go
@@ -6,22 +6,22 @@ package mgostorage
 import (
 	"gopkg.in/errgo.v1"
 	"gopkg.in/mgo.v2"
-	"gopkg.in/mgo.v2/bson"
 
 	"gopkg.in/macaroon-bakery.v1/bakery"
 )
 
 // New returns an implementation of Storage
 // that stores all items in MongoDB.
+// It never returns an error (the error return
+// is for backward compatibility with a previous
+// version that could return an error).
+//
+// Note that the caller is responsible for closing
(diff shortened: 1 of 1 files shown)


mhilton:
:+1: LGTM

bakery/mgostorage/storage.go
```diff
@@ -6,21 +6,22 @@ package mgostorage
...
+// the mgo session associated with the collection.
 func New(c *mgo.Collection) (bakery.Storage, error) {
 	m := mgoStorage{
 		col: c,
 	}
-	err := m.setUpCollection()
-	if err != nil {
-		return nil, errgo.Mask(err)
-	}
 	return &m, nil
```
alesstimec: return &mgoStorage{col:c}, nil

bakery/mgostorage/storage.go
```diff
@@ -29,43 +30,25 @@ type mgoStorage struct {
 }
 
 type storageDoc struct {
-	Location string `bson:"loc"`
+	Location string `bson:"_id"`
 	Item     string `bson:"item"`
-}
 
-func (s *mgoStorage) setUpCollection() error {
```
alesstimec: Does this mean the caller is now responsible for setting up the collection index? I'm not completely sure i like this..

alesstimec:
LGTM with a nitpick

bakery/mgostorage/storage.go
rogpeppe: done, slightly differently (no need to allocate a pointer)

PR #53, opened 2015-07-10
add support for agent login mechanism

mhilton:
Add the ability to access the visit URL as an agent and perform login interactions without opening a browser.

### DIFF
--- a/httpbakery/agent/agent.go
+++ b/httpbakery/agent/agent.go
@@ -1,0 +1,197 @@
+// Package agent enables non-interactive (agent) login using macaroons.
+// To enable agent authorization with a given httpbakery.Client c against
+// a given third party discharge server URL u:
+//
+// 	SetUpAuth(c, u, agentUsername)
+//
+package agent
+
+import (
+	"bytes"
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"mime"
+	"net/http"
+	"net/url"
+
+	"github.com/juju/loggo"
+	"gopkg.in/errgo.v1"
+
+	"gopkg.in/macaroon-bakery.v1/bakery"
+	"gopkg.in/macaroon-bakery.v1/httpbakery"
+)
+
+var logger = loggo.GetLogger("httpbakery.agent")
+
+/*
+PROTOCOL
+
+An agent login works as follows:
+
+	    Agent                            Login Service
+	      |                                    |
+	      | GET visitURL with agent cookie     |
+	      |----------------------------------->|
+	      |                                    |
+	      |    Macaroon with local third-party |
+	      |                             caveat |
+	      |<-----------------------------------|
+	      |                                    |
+	      | GET visitURL with agent cookie &   |
+	      | discharged macaroon                |
+	      |----------------------------------->|
+	      |                                    |
+	      |               Agent login response |
+	      |<-----------------------------------|
+	      |                                    |
+
+The agent cookie is a cookie named "agent-login" holding a base64
+encoded JSON object of the form:
+
+	{
+		"username": <username>,
+		"public_key": <public key>
+	}
+
+A local third-party caveat is a third party caveat with the location
+set to "local" and the caveat encrypted with the public key declared
+in the agent cookie. The httpbakery.Client automatically discharges
+the local third-party caveat.
+
+On success the response is a JSON object of the form:
+
+	{
+		"agent_login": true
+	}
+
+If an error occurs then the response should be a JSON object that
+unmarshalls to an httpbakery.Error.
+*/
+
+// agentLogin defines the structure of an agent login cookie. It is also
+// returned in a successful agent login attempt to help indicate that an
+// agent login has occurred.
+type agentLogin struct {
+	Username  string            `json:"username"`
+	PublicKey *bakery.PublicKey `json:"public_key"`
+}
+
+// agentResponse defines the structure of an agent login response.
+type agentResponse struct {
+	AgentLogin bool `json:"agent_login"`
+}
+
+// SetUpAuth configures agent authentication on c. A cookie is created in
+// c's cookie jar containing credentials derived from the username and
+// c.Key. c.VisitWebPage is set to VisitWebPage(c). The return is
+// non-nil only if c.Key is nil.
+func SetUpAuth(c *httpbakery.Client, u *url.URL, username string) error {
+	if c.Key == nil {
+		return errgo.New("cannot set-up authentication: client key not configured")
+	}
+	SetCookie(c.Jar, u, username, &c.Key.Public)
+	c.VisitWebPage = VisitWebPage(c)
+	return nil
+}
+
+// SetCookie creates a cookie in jar which is suitable for performing agent
+// logins to u.
+func SetCookie(jar http.CookieJar, u *url.URL, username string, pk *bakery.PublicKey) {
+	al := agentLogin{
+		Username:  username,
+		PublicKey: pk,
+	}
+	b, err := json.Marshal(al)
+	if err != nil {
+		// This shouldn't happen as the agentLogin type has to be marshalable.
+		panic(errgo.Notef(err, "cannot marshal cookie"))
+	}
+	v := base64.StdEncoding.EncodeToString(b)
+	jar.SetCookies(u, []*http.Cookie{{
+		Name:  "agent-login",
+		Value: v,
+	}})
+}
+
+// VisitWebPage creates a function that can be used with
+// httpbakery.Client.VisitWebPage. The function uses c to access the
+// visit URL. If no agent-login cookie has been configured for u an error
+// with the cause of http.ErrNoCookie will be returned. If the login
+// fails the returned error will be of type *httpbakery.Error. If the
+// response from the visitURL cannot be interpreted the error will be of
+// type *UnexpectedResponseError.
+func VisitWebPage(c *httpbakery.Client) func(u *url.URL) error {
+	return func(u *url.URL) error {
+		err := http.ErrNoCookie
+		for _, c := range c.Jar.Cookies(u) {
+			if c.Name == "agent-login" {
+				err = nil
+				break
+			}
+		}
+		if err != nil {
+			return errgo.WithCausef(err, http.ErrNoCookie, "cannot perform agent login")
+		}
+		req, err := http.NewRequest("GET", u.String(), nil)
+		if err != nil {
+			return errgo.Notef(err, "cannot create request")
+		}
+		resp, err := c.Do(req)
+		if err != nil {
+			return errgo.Notef(err, "cannot perform request")
+		}
+		defer resp.Body.Close()
+		b, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			logger.Errorf("cannot read response body: %s", err)
+			b = []byte{}
+		}
+		mt, _, err := mime.ParseMediaType(resp.Header.Get("Content-Type"))
+		if err != nil {
+			logger.Warningf("cannot parse response content type: %s", err)
+			mt = ""
+		}
+		if mt != "application/json" {
+			uerr := (*UnexpectedResponseError)(resp)
+			uerr.Body = ioutil.NopCloser(bytes.NewReader(b))
+			return uerr
+		}
+		if resp.StatusCode != http.StatusOK {
+			var herr httpbakery.Error
+			err := json.Unmarshal(b, &herr)
+			if err == nil && herr.Message != "" {
+				return &herr
+			}
+			if err != nil {
+				logger.Warningf("cannot unmarshal error response: %s", err)
+			}
+			uerr := (*UnexpectedResponseError)(resp)
+			uerr.Body = ioutil.NopCloser(bytes.NewReader(b))
+			return uerr
+		}
+		var ar agentResponse
+		err = json.Unmarshal(b, &ar)
+		if err == nil && ar.AgentLogin {
+			return nil
+		}
+		if err != nil {
+			logger.Warningf("cannot unmarshal response: %s", err)
+		}
+		uerr := (*UnexpectedResponseError)(resp)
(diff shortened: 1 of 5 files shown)


httpbakery/agent/agent.go
```diff
@@ -0,0 +1,154 @@
...
+// an agent.
+func SetCookie(c *httpbakery.Client, u *url.URL, username string) {
+	al := agentLogin{
+		Username:  username,
+		PublicKey: &c.Key.Public,
+	}
+	b, err := json.Marshal(al)
+	if err != nil {
+		// This shouldn't happen as the agentLogin type has to be marshalable.
+		panic(errgo.Notef(err, "cannot marshal cookie"))
```
alesstimec, line 109: I agree, but i'm not the biggest fan of this approach..

httpbakery/agent/agent.go
```diff
@@ -0,0 +1,154 @@
...
+			return errgo.Notef(err, "agent login failed")
+		}
+		return nil
+	}
+}
+
+// getError checks for an error in r. If the status code indicates there
+// is an error then getError attempts to unmarshal the error as an
+// httpbakery.Error. If it is unmarshalled succesfully then this error is
+// returned, otherwise an error based on the respnse status is returned.
```
alesstimec: s/respnse/response/

httpbakery/agent/agent.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+	v := base64.StdEncoding.EncodeToString(b)
+	c.Jar.SetCookies(u, []*http.Cookie{{
+		Name:  "agent-login",
+		Value: v,
+	}})
+}
+
+// VisitWebPage creates a function that can be used with
+// httpbaker.Client.VisitWebPage. The function uses c to access the visit
```
alesstimec: s/httpbaker/httpbakery/

httpbakery/agent/agent.go
```diff
@@ -0,0 +1,154 @@
...
+	"github.com/juju/loggo"
+	"gopkg.in/errgo.v1"
+
+	"gopkg.in/macaroon-bakery.v1/bakery"
+	"gopkg.in/macaroon-bakery.v1/httpbakery"
+)
+
+var logger = loggo.GetLogger("httpbakery.agent")
+
+// agentLogin contains the structure of an agent login cooke. It is also
```
alesstimec: agentLogin defines

httpbakery/agent/agent.go
mhilton: done

httpbakery/agent/agent.go
mhilton: done

alesstimec:
LGTM with a few nitpicks..

httpbakery/agent/agent.go
mhilton: done

httpbakery/agent/agent_test.go
```diff
@@ -0,0 +1,248 @@
...
+		err = s.bakery.Check(ms, bakery.FirstPartyCheckerFunc(
+			func(caveat string) error {
+				return nil
+			},
+		))
+		c.Assert(err, gc.IsNil)
+	}
+}
+
+func ExampleVisitWebPage() {
```
frankban, line 139: Is this used?

httpbakery/agent/agent_test.go
```diff
@@ -0,0 +1,248 @@
...
+	u, err := url.Parse(s.discharger.URL)
+	c.Assert(err, gc.IsNil)
+	for i, test := range agentLoginTests {
+		c.Logf("%d. %s", i, test.about)
+		s.discharger.LoginHandler = test.loginHandler
+		client := httpbakery.NewClient()
+		client.Key, err = bakery.GenerateKey()
+		c.Assert(err, gc.IsNil)
+		agent.SetCookie(client, u, "test-user")
+		client.VisitWebPage = agent.VisitWebPage(client)
```
frankban: This seems a little weird, but I think I understand why it's required.

frankban:
:+1: ty!

httpbakery/agent/agent.go
```diff
@@ -0,0 +1,161 @@
...
+	      |                                    |
+	      | GET visitURL with agent cookie &   |
+	      | discharged macaroon                |
+	      |----------------------------------->|
+	      |                                    |
+	      |               Agent login response |
+	      |<-----------------------------------|
+	      |                                    |
+
+The agent cookie is a cookie named "agent-login" its value is a base64
```
rogpeppe: The agent cookie is a cookie named "agent-login" holding a base64-encoded JSON object of the form:

?

httpbakery/agent/agent.go
rogpeppe, line 109: Better to do this than to force each caller to check for an error that
can actually never happen, I think.

httpbakery/agent/agent.go
```diff
@@ -0,0 +1,161 @@
...
+
+// SetCookie creates a cookie in c which is suitable for performing agent
+// logins to u. The username will be combined with the public key taken
+// from c.Key.Public, c.Key must have been set before calling SetCookie.
+// If c is used with VisitWebPage interactions with u will be performed as
+// an agent.
+func SetCookie(c *httpbakery.Client, u *url.URL, username string) {
+	al := agentLogin{
+		Username:  username,
+		PublicKey: &c.Key.Public,
```
rogpeppe: deliberate panic if c.Key is nil. or add an error return for this
case, which may well be not that uncommon as bakery clients
are often passed in from outside so quite easy to make the mistake.

httpbakery/agent/agent.go
```diff
@@ -0,0 +1,161 @@
...
+		Value: v,
+	}})
+}
+
+// VisitWebPage creates a function that can be used with
+// httpbakery.Client.VisitWebPage. The function uses c to access the
+// visit URL which should previously have been configured with SetCookie.
+func VisitWebPage(c *httpbakery.Client) func(u *url.URL) error {
+	return func(u *url.URL) error {
+		req, err := http.NewRequest("GET", u.String(), nil)
```
rogpeppe, line 137: Shouldn't we have an Accept: application/json header for this request?

mhilton, line 137: No, I think that becomes unnecessary here. Because we aren't querying the login methods at this point, there is no real need to.

httpbakery/agent/agent.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+// getError checks for an error in r. If the status code indicates there
+// is an error then getError attempts to unmarshal the error as an
+// httpbakery.Error. If it is unmarshalled succesfully then this error is
+// returned, otherwise an error based on the response status is returned.
+// getError returns nil only when the status code indicates there was no
+// error.
+func getError(r *http.Response) error {
+	if r.StatusCode < 400 {
```
rogpeppe: Do we really want to try to unmarshal the body if we've
got a 30x error?

httpbakery/agent/agent.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+		resp, err := c.Do(req)
+		if err != nil {
+			return errgo.Notef(err, "cannot perform request")
+		}
+		defer resp.Body.Close()
+		if err := getError(resp); err != nil {
+			return errgo.Notef(err, "cannot log in")
+		}
+		var al agentLogin
```
rogpeppe: It seems somewhat odd that the response type is the same
as the cookie type and just confirms information that we
already know. At the least, this response type and its
expected values should be documented in the protocol description.
Probably just checking for a 200 response is sufficient.

httpbakery/agent/agent.go
rogpeppe: Alternatively make the args to SetCookie a little closer
to what it actually needs:

```
func SetCookie(u *url.URL, username string, key *bakery.PublicKey)
```

then it's obvious to the caller that the public key is required,
and perhaps the role of the function is a little clearer too.

httpbakery/agent/agent.go
mhilton: That's a good point, In practice it doesn't matter much as it will likely fail at the next hurdle, but I'll change it.

httpbakery/agent/agent.go
mhilton: The reason for doing this is to be able to tell that the thing that succeeded was what we expected to succeed. The problem is that if there is a visitURL that doesn't understand the cookie will most likely return a login form with a corresponding 200 status. At which point this function thinks we've successfully logged in, but actually we are waiting for an interaction to happen.

Maybe a 204 (No Content) would be better that would at least make it much less likely that any further interaction is expected by either end without returning unnecessary data.

httpbakery/agent/agent.go
mhilton: done

httpbakery/agent/agent.go
mhilton: done.

httpbakery/agent/agent.go
mhilton: I've changed this to only accept a 204 as success. That should be enough to eliminate any practical false positives.

httpbakery/agent/agent.go
```diff
@@ -0,0 +1,150 @@
+/*
+Package agent implements the agent login mechanism. The agent login
+mechanism provides a form of interaction that can be used in
+situations where interaction with the user is not possible.
+
+An agent login works as follows:
```
rogpeppe, line 31: Most of this comment is actually not necessary if you just want to use this package.
I wonder if most of this comment (useful as it is) should perhaps be an internal
comment, and this package doc should just outline how to use the package.

e.g.

```
Package agent enables non-interactive (agent) login using macaroons.
To enable agent authorization with a given httpbakery.Client c against
a given third party discharge server URL u:

    agent.SetCookie(c, u, agentUsername, &c.Key.Public)
    c.VisitWebPage = agent.VisitWebPage(c)
```

writing this down like this makes me think that perhaps the API
is more complex than it needs to be.

How about:

```
// SetUpAuth enables non-interactive authorization on the given client.
// Whenever the client is directed to obtain a discharge macaroon
// from a 3rd party under the given URL, the given username
// will be provided along with the client's public key.
// SetUpAuth does this by setting a cookie in c.Client.Jar and
// setting c.VisitWebPage to a direct web page fetcher.
func SetUpAuth(c *httpbakery.Client, u *url.URL, username string)
```

 ?

httpbakery/agent/agent.go
```diff
@@ -0,0 +1,181 @@
...
+	v := base64.StdEncoding.EncodeToString(b)
+	jar.SetCookies(u, []*http.Cookie{{
+		Name:  "agent-login",
+		Value: v,
+	}})
+}
+
+// VisitWebPage creates a function that can be used with
+// httpbakery.Client.VisitWebPage. The function uses c to access the
+// visit URL which should previously have been configured with SetCookie.
```
rogpeppe: Whether the visit URL is or is not configured with SetCookie is not under the
control of the caller - it depends on the macaroon being discharged.

That makes me wonder if the function should check that there is
actually an agent-login cookie registered for the given URL before
doing the visit. What do you think?

httpbakery/agent/agent.go
```diff
@@ -0,0 +1,181 @@
...
+// If the login fails the returned error will be of type
+// *httpbakery.Error. If the response from the visitURL cannot be
+// interpretted the error will be of type *UnexpectedResponseError.
```
rogpeppe: s/prett/pret/

mhilton: I thought that looked wrong. I've done it somewhere else too.

httpbakery/agent/agent.go
```diff
@@ -0,0 +1,181 @@
...
+		if err != nil {
+			logger.Errorf("cannot read response body: %s", err)
+			b = []byte{}
+		}
+		mt, _, err := mime.ParseMediaType(resp.Header.Get("Content-Type"))
+		if err != nil {
+			logger.Warningf("cannot parse response content type: %s", err)
+			mt = ""
+		}
+		if mt == "application/json" {
```
rogpeppe:
```
if mt != "application/json" {
   return ...
}
```

saving on indentation?

httpbakery/agent/agent.go
```diff
@@ -0,0 +1,181 @@
...
+
+func (u *UnexpectedResponseError) Error() string {
+	return fmt.Sprintf(
+		"unexpected response to non-interactive web page visit %s (content type %s)",
+		u.Request.URL.String(),
+		u.Header.Get("Content-Type"))
+}
+
+/*
+PROTOCOL
```
rogpeppe: Put this comment near the start of the file, just before type agentLogin, perhaps?

httpbakery/agent/discharge_test.go
```diff
@@ -0,0 +1,181 @@
...
+	"gopkg.in/macaroon-bakery.v1/httpbakery"
+	"gopkg.in/macaroon-bakery.v1/httpbakery/agent"
+)
+
+type discharge struct {
+	cavId string
+	c     chan error
+}
+
+type Discharger struct {
```
rogpeppe, line 26: Thought for the future: it might be nice if the bakerytest discharger could support this kind of flexibility.

rogpeppe:
LGTM with some final thoughts. Thanks!

httpbakery/agent/agent.go
mhilton: I have made it do this and quite like it.

httpbakery/agent/agent.go
mhilton: done

httpbakery/agent/agent.go
mhilton: It did increase some other code duplication, but it's probably nicer overall.

httpbakery/agent/agent.go
mhilton: done

PR #55, opened 2015-07-23
httpbakery: add some more client tests

rogpeppe:
The error message produced on a wait failure was uninformative.
This fixes that and adds some more tests. More can be added
to the table later, but this substantially increases test coverage as is.

### DIFF
--- a/httpbakery/client.go
+++ b/httpbakery/client.go
@@ -428,9 +428,10 @@ func (c *Client) interact(location, visitURLStr, waitURLStr string) (*macaroon.M
 		if err := json.NewDecoder(waitResp.Body).Decode(&resp); err != nil {
 			return nil, errgo.Notef(err, "cannot unmarshal wait error response")
 		}
-		return nil, errgo.WithCausef(nil, &DischargeError{
+		dischargeErr := &DischargeError{
 			Reason: &resp,
-		}, "failed to acquire macaroon after waiting")
+		}
+		return nil, errgo.NoteMask(dischargeErr, "failed to acquire macaroon after waiting", errgo.Any)
 	}
 	var resp WaitResponse
 	if err := json.NewDecoder(waitResp.Body).Decode(&resp); err != nil {
--- a/httpbakery/client_test.go
+++ b/httpbakery/client_test.go
@@ -325,6 +325,101 @@ func (s *ClientSuite) TestDischargeWithInteractionRequiredError(c *gc.C) {
 	c.Assert(resp, gc.IsNil)
 }
 
+var dischargeWithVisitURLErrorTests = []struct {
+	about       string
+	respond     func(http.ResponseWriter)
+	expectError string
+}{{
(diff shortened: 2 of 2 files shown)


httpbakery/client.go
```diff
@@ -428,9 +428,10 @@ func (c *Client) interact(location, visitURLStr, waitURLStr string) (*macaroon.M
 		if err := json.NewDecoder(waitResp.Body).Decode(&resp); err != nil {
 			return nil, errgo.Notef(err, "cannot unmarshal wait error response")
 		}
-		return nil, errgo.WithCausef(nil, &DischargeError{
+		dischargeErr := &DischargeError{
```
mhilton, line 431: As this could also be a message from apache it would be really helpful if it somehow included the body text if the response isn't some JSON.

httpbakery/client_test.go
```diff
@@ -325,6 +325,108 @@ func (s *ClientSuite) TestDischargeWithInteractionRequiredError(c *gc.C) {
...
+	respond     func(http.ResponseWriter)
+	expectError string
+}{{
+	about: "error message",
+	respond: func(w http.ResponseWriter) {
+		jsonhttp.WriteError(httpbakery.ErrorToResponse)(w, fmt.Errorf("an error"))
+	},
+	expectError: `cannot get discharge from ".*": failed to acquire macaroon after waiting: third party refused discharge: an error`,
+}, {
+	about: "non-JSON error",
```
mhilton, line 339: This non-JSON looks a lot like JSON.

frankban, line 339: Seems like a repeated test here.

httpbakery/client_test.go
```diff
@@ -325,6 +325,108 @@ func (s *ClientSuite) TestDischargeWithInteractionRequiredError(c *gc.C) {
...
+		respond: respond,
+		mux:     http.NewServeMux(),
+	}
+	h.mux.HandleFunc("/visit", h.serveVisit)
+	h.mux.HandleFunc("/wait", h.serveWait)
+	return h
+}
+
+func (h *visitHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+	h.mux.ServeHTTP(w, req)
```
mhilton, line 411: why not just embed a mux?

frankban:
:+1: with one minor change and thanks for the tests, but does this actually improve error messages?

mhilton:
LGTM modulo a couple of comments.

httpbakery/client.go
rogpeppe, line 431: Yes, I plan to do that in a future PR, probably by implementing it in httprequest and using that in httpbakery.

httpbakery/client_test.go
rogpeppe, line 411: Because I don't like embedding-for-convenience particularly - we don't want to expose any of the other mux methods.

PR #57, opened 2015-08-31
bakerytest: always use TLS

rogpeppe:
This means it's OK for production code to insist on using an HTTPS
URL for an identity location.

### DIFF
--- a/bakerytest/bakerytest.go
+++ b/bakerytest/bakerytest.go
@@ -3,8 +3,10 @@
 package bakerytest
 
 import (
+	"crypto/tls"
 	"net/http"
 	"net/http/httptest"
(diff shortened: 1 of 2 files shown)


alesstimec:
LGTM

bakerytest/bakerytest.go
```diff
@@ -21,6 +23,51 @@ type Discharger struct {
 	server *httptest.Server
 }
 
+var skipVerify struct {
+	mu            sync.Mutex
```
mhilton, line 27: you could possibly use sync/atomic instead  of a lock here , but  it isn't important.

mhilton:
LGTM

PR #59, opened 2015-09-03
Replace github.com/juju/utils/jsonhttp with github.com/juju/httprequest.

tasdomas:


bakery/example/idservice/idservice.go
```diff
@@ -59,25 +59,25 @@ func New(p Params) (http.Handler, error) {
 	}
 	mux := http.NewServeMux()
```
rogpeppe, line 61: I think perhaps this could use an httprouter mux. We couldn't do that in httpbakery because the
http mux is part of the public API but I think it would make sense here.

httpbakery/discharge.go
```diff
@@ -123,3 +125,10 @@ func randomBytes(n int) ([]byte, error) {
 	}
 	return b, nil
 }
+
+// MkHandler converts an httprouter.Handle into an http.Handler.
```
rogpeppe: This doesn't belong in httpbakery - its use of httprouter types is entirely an internal implementation thing. If you use httprouter in the examples, you won't need this there anyway, I think.

rogpeppe:
Looks great, but I'm not that happy about exporting MkHandler from httpbakery. Perhaps just copy/paste the function and leave a TODO in the id server example?

rogpeppe:
LGTM, thanks!

PR #60, opened 2015-09-09
httpbakery: return status 401 discharge-required errors

rogpeppe:
We preserve backward compatibility. Servers will need to use the new *ForRequest
functions to serve the new protocol.


makyo:
:+1: Looks like a good path forward

fabricematrat:
:+1:

httpbakery/error.go
```diff
@@ -38,8 +39,23 @@ type Error struct {
 	Code    ErrorCode  `json:",omitempty"`
 	Message string     `json:",omitempty"`
 	Info    *ErrorInfo `json:",omitempty"`
+
+	// version holds the protocol version that was used
+	// to create the error (see NewDischargeRequiredErrorWithVersion).
+	version version
 }
 
+// version represents a version of the bakery protocol. It is jused
```
mhilton, line 48: *jused

httpbakery/error.go
```diff
@@ -38,8 +39,23 @@ type Error struct {
...
+// to determine the kind of response to send when there is a
+// discharge-required error.
+type version int
+
+const (
+	version0      version = 0
```
mhilton, line 54: do these constants really add anything?

httpbakery/error.go
```diff
@@ -91,14 +107,30 @@ func (e *Error) ErrorInfo() *ErrorInfo {
...
 func ErrorToResponse(err error) (int, interface{}) {
 	errorBody := errorResponseBody(err)
+	var body interface{} = errorBody
 	status := http.StatusInternalServerError
 	switch errorBody.Code {
 	case ErrBadRequest:
 		status = http.StatusBadRequest
 	case ErrDischargeRequired, ErrInteractionRequired:
-		status = http.StatusProxyAuthRequired
+		switch errorBody.version {
```
mhilton: I might be wrong but I can't see where errorBody.version is set. won't this always be 0?

mhilton: ignore that I worked it out.

mhilton:
LGTM with some minor thoughts.

httpbakery/error.go
rogpeppe, line 54: I went back and forth on this. I ended up thinking they are useful because they enumerate all the versions that we explicitly support.

PR #61, opened 2015-09-15
httpbakery: add NewInteractionRequiredError function

rogpeppe:
The discharge-required error isn't the only one that needs special treatment.
This allows the interaction-required error to return a 401 response too,
while preserving backward compatibility.


mhilton:
:+1: LGTM

fabricematrat:
:+1:

PR #62, opened 2015-09-16
Support form visit web page protocol

mhilton:
Also add a LoginDischarger to bakerytest.

### DIFF
--- a/bakerytest/bakerytest.go
+++ b/bakerytest/bakerytest.go
@@ -4,10 +4,14 @@ package bakerytest
 
 import (
 	"crypto/tls"
+	"fmt"
 	"net/http"
 	"net/http/httptest"
 	"sync"
+	"time"
 
+	"github.com/juju/httprequest"
+	"gopkg.in/errgo.v1"
 	"gopkg.in/macaroon-bakery.v1/bakery"
 	"gopkg.in/macaroon-bakery.v1/bakery/checkers"
 	"gopkg.in/macaroon-bakery.v1/httpbakery"
@@ -126,4 +130,175 @@ func (d *Discharger) PublicKeyForLocation(loc string) (*bakery.PublicKey, error)
 		return d.Service.PublicKey(), nil
 	}
 	return nil, bakery.ErrNotFound
+}
+
+type discharge struct {
+	cavId string
+	c     chan error
+}
+
+// InteractiveDischarger is a Discharger that always requires interraction to
+// complete the discharge.
+type InteractiveDischarger struct {
+	Discharger
+	Mux *http.ServeMux
+
+	// mu protects the following fields.
+	mu      sync.Mutex
+	waiting map[string]discharge
+	id      int
+}
+
+// NewInteractiveDischarger returns a new InteractiveDischarger. The
+// InteractiveDischarger will serve the following endpoints by default:
+//
+//     /discharge - always causes interaction to be required.
+//     /publickey - gets the bakery public key.
+//     /visit - delegates to visitHandler.
+//     /wait - blocks waiting for the interaction to complete.
+//
+// Additional endpoints may be added to Mux as necessary.
+//
+// The /discharge endpoint generates a error with the code
+// httpbakery.ErrInterractionRequired. The visitURL and waitURL will
+// point to the /visit and /wait endpoints of the InteractiveDischarger
+// respectively. These URLs will also carry context information in query
+// parameters, any handlers should be careful to preserve this context
+// information between calls. The easiest way to do this is to always use
+// the URL method when generating new URLs.
+//
+// The /visit endpoint is handled by the provided visitHandler. This
+// handler performs the required interactions and should result in the
+// FinishInteraction method being called. This handler may process the
+// interaction in a number of steps, possibly using additional handlers,
+// so long as FinishInteraction is called when no further interaction is
+// required.
+//
+// The /wait endpoint blocks until FinishInteraction has been called by
+// the corresponding /visit endpoint, or another endpoint triggered by
+// visitHandler.
+//
+// If locator is non-nil, it will be used to find public keys
+// for any third party caveats returned by the checker.
+//
+// Calling this function has the side-effect of setting
+// InsecureSkipVerify in http.DefaultTransport.TLSClientConfig
+// until all the dischargers are closed.
+//
+// The returned InteractiveDischarger must be closed when finished with.
+func NewInteractiveDischarger(locator bakery.PublicKeyLocator, visitHandler http.Handler) *InteractiveDischarger {
+	d := &InteractiveDischarger{
+		Mux:     http.NewServeMux(),
+		waiting: map[string]discharge{},
+	}
+	d.Mux.Handle("/visit", visitHandler)
+	d.Mux.Handle("/wait", http.HandlerFunc(d.wait))
+	server := httptest.NewTLSServer(d.Mux)
+	svc, err := bakery.NewService(bakery.NewServiceParams{
+		Location: server.URL,
+		Locator:  locator,
+	})
+	if err != nil {
+		panic(err)
+	}
+	httpbakery.AddDischargeHandler(d.Mux, "/", svc, d.checker)
+	startSkipVerify()
+	d.Discharger = Discharger{
+		Service: svc,
+		server:  server,
+	}
+	return d
+}
+
+func (d *InteractiveDischarger) checker(req *http.Request, cavId, cav string) ([]checkers.Caveat, error) {
+	d.mu.Lock()
+	id := fmt.Sprintf("%d", d.id)
+	d.id++
+	d.waiting[id] = discharge{cavId, make(chan error, 1)}
+	d.mu.Unlock()
+	visitURL := fmt.Sprintf("%s/visit?waitid=%s", d.Discharger.server.URL, id)
+	waitURL := fmt.Sprintf("%s/wait?waitid=%s", d.Discharger.server.URL, id)
+	return nil, httpbakery.NewInteractionRequiredError(visitURL, waitURL, nil, req)
+}
+
+func (d *InteractiveDischarger) wait(w http.ResponseWriter, r *http.Request) {
+	r.ParseForm()
+	d.mu.Lock()
+	discharge, ok := d.waiting[r.Form.Get("waitid")]
+	d.mu.Unlock()
+	if !ok {
+		code, body := httpbakery.ErrorToResponse(errgo.Newf("invalid waitid %q", r.Form.Get("waitid")))
+		httprequest.WriteJSON(w, code, body)
+		return
+	}
+	defer func() {
+		d.mu.Lock()
+		delete(d.waiting, r.Form.Get("waitid"))
+		d.mu.Unlock()
+	}()
+	var err error
+	select {
+	case e := <-discharge.c:
+		err = e
+	case <-time.After(5 * time.Minute):
+		code, body := httpbakery.ErrorToResponse(errgo.New("timeout waiting for interaction to complete"))
+		httprequest.WriteJSON(w, code, body)
+		return
+	}
+	if err != nil {
+		code, body := httpbakery.ErrorToResponse(err)
+		httprequest.WriteJSON(w, code, body)
+		return
+	}
+	m, err := d.Service.Discharge(
+		bakery.ThirdPartyCheckerFunc(
+			func(cavId, caveat string) ([]checkers.Caveat, error) {
+				return nil, nil
+			},
+		),
+		discharge.cavId,
+	)
+	if err != nil {
+		code, body := httpbakery.ErrorToResponse(err)
+		httprequest.WriteJSON(w, code, body)
+		return
+	}
+	httprequest.WriteJSON(
+		w,
+		http.StatusOK,
+		httpbakery.WaitResponse{
+			Macaroon: m,
+		},
+	)
+}
+
+// FinishInteraction signals to the InteractiveDischarger that a
+// particular interaction is complete. It causes any waiting requests to
+// return. If err is not nil then it will be returned by the
+// corresponding /wait request.
+func (d *InteractiveDischarger) FinishInteraction(w http.ResponseWriter, r *http.Request, err error) {
+	r.ParseForm()
+	d.mu.Lock()
+	discharge, ok := d.waiting[r.Form.Get("waitid")]
+	d.mu.Unlock()
+	if !ok {
+		code, body := httpbakery.ErrorToResponse(errgo.Newf("invalid waitid %q", r.Form.Get("waitid")))
+		httprequest.WriteJSON(w, code, body)
+		return
+	}
+	select {
+	case discharge.c <- err:
+	default:
+		panic("cannot finish interaction " + r.Form.Get("waitid"))
+	}
+	return
+}
+
+// URL returns a URL addressed to the given path in the discharger that
+// contains any discharger context information found in the given
+// request. Use this to generate intermediate URLs before calling
+// FinishInteraction.
+func (d *InteractiveDischarger) URL(path string, r *http.Request) string {
+	r.ParseForm()
+	return d.Location() + path + "?waitid=" + r.Form.Get("waitid")
 }
--- a/httpbakery/form/form.go
+++ b/httpbakery/form/form.go
@@ -1,0 +1,151 @@
+// Package form enables interactive login without using a web browser.
+package form
+
+import (
+	"net/http"
+	"net/url"
+
+	"github.com/juju/httprequest"
+	"gopkg.in/errgo.v1"
+	"gopkg.in/juju/environschema.v1"
+
+	"gopkg.in/macaroon-bakery.v1/httpbakery"
+)
+
+/*
+PROTOCOL
+
+A form login works as follows:
+
+	   Client                            Login Service
+	      |                                    |
+	      | GET visitURL with                  |
+	      | "Accept: application/json"         |
+	      |----------------------------------->|
+	      |                                    |
+	      |   Login Methods (including "form") |
+	      |<-----------------------------------|
+	      |                                    |
+	      | GET "form" URL                     |
+	      |----------------------------------->|
+	      |                                    |
+	      |                  Schema definition |
+	      |<-----------------------------------|
+	      |                                    |
+	+-------------+                            |
+	|   Client    |                            |
+	| Interaction |                            |
+	+-------------+                            |
+	      |                                    |
+	      | POST data to "form" URL            |
+	      |----------------------------------->|
+	      |                                    |
+	      |                Form login response |
+	      |<-----------------------------------|
+	      |                                    |
+
+The schema is provided as a environschema.Fileds object. It is the
+client's responsibility to interpret the schema and present it to the
+user.
+*/
+
+// Filler represents an object that can fill out a form. The given schema
+// represents the form template. The returned value should be compatible
+// with this.
+type Filler interface {
+	Fill(schema environschema.Fields) (map[string]interface{}, error)
+}
+
+// SetUpAuth configures form authentication on c. The VisitWebPage field
+// in c will be set to a function that will attempt form based
+// authentication using h to perform the interaction with the user.
+//
+// The VisitWebPage function downloads the schema from the specified
+// server and calls h.Handle. The map returned by h.Handle should match
+// the schema specified, but VisitWebPage will not verify this before
+// sending to the server. Any errors returned by h.Handle will not have
+// their cause masked.
+func SetUpAuth(c *httpbakery.Client, f Filler) {
+	v := webPageVisitor{
+		client: &httprequest.Client{
+			Doer: c,
+		},
+		filler: f,
+	}
+	c.VisitWebPage = v.visitWebPage
+}
+
+// webPageVisitor contains the state required by visitWebPage.
+type webPageVisitor struct {
+	client *httprequest.Client
+	filler Filler
+}
+
+// loginMethods contains the response expected from the login URL. It
+// only checks for the "form" method as that is the only one that can be
+// handled.
+type loginMethods struct {
+	Form string `json:"form"`
+}
+
+// SchemaRequest is a request for a form schema.
+type SchemaRequest struct {
+	httprequest.Route `httprequest:"GET"`
+}
+
+// SchemaResponse contains the message expected in response to the schema
+// request.
+type SchemaResponse struct {
+	Schema environschema.Fields `json:"schema"`
+}
+
+// LoginRequest is a request to perform a login using the provided details.
+type LoginRequest struct {
+	httprequest.Route `httprequest:"POST"`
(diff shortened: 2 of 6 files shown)


bakerytest/bakerytest.go
```diff
@@ -127,3 +130,131 @@ func (d *Discharger) PublicKeyForLocation(loc string) (*bakery.PublicKey, error)
...
+	Mux *http.ServeMux
+
+	// mu protects the following fields.
+	mu      sync.Mutex
+	waiting map[string]discharge
+	id      int
+}
+
+// NewLoginDischarger returns a new LoginDischarger which uses
+// loginHandler to handle the login interraction. The login is completed
```
rogpeppe: s/interraction/interaction/

rogpeppe: I think we should be more specific about what "handling the login interaction" actually means - for example it should perhaps mention that it is called on /login and (I presume) it's the responsibility of the handler to call FinishLogin.

bakerytest/bakerytest.go
```diff
@@ -127,3 +130,131 @@ func (d *Discharger) PublicKeyForLocation(loc string) (*bakery.PublicKey, error)
...
+// by calling FinishLogin.
+//
+// If locator is non-nil, it will be used to find public keys
+// for any third party caveats returned by the checker.
+//
+// Calling this function has the side-effect of setting
+// InsecureSkipVerify in http.DefaultTransport.TLSClientConfig
+// until all the dischargers are closed.
+func NewLoginDischarger(locator bakery.PublicKeyLocator, loginHandler http.Handler) *LoginDischarger {
+	ld := &LoginDischarger{
```
rogpeppe: I'd be tempted just to use "d".

bakerytest/bakerytest.go
```diff
@@ -127,3 +130,131 @@ func (d *Discharger) PublicKeyForLocation(loc string) (*bakery.PublicKey, error)
...
+// NewLoginDischarger returns a new LoginDischarger which uses
+// loginHandler to handle the login interraction. The login is completed
+// by calling FinishLogin.
+//
+// If locator is non-nil, it will be used to find public keys
+// for any third party caveats returned by the checker.
+//
+// Calling this function has the side-effect of setting
+// InsecureSkipVerify in http.DefaultTransport.TLSClientConfig
+// until all the dischargers are closed.
```
rogpeppe, line 186: // The returned discharger must be closed when finished with.

?

bakerytest/bakerytest.go
```diff
@@ -127,3 +130,131 @@ func (d *Discharger) PublicKeyForLocation(loc string) (*bakery.PublicKey, error)
...
+	ld.Discharger = Discharger{
+		Service: svc,
+		server:  server,
+	}
+	return ld
+}
+
+func (d *LoginDischarger) checker(req *http.Request, cavId, cav string) ([]checkers.Caveat, error) {
+	d.mu.Lock()
+	id := fmt.Sprintf("%04x", d.id)
```
rogpeppe: Why not just %d ?

bakerytest/bakerytest.go
```diff
@@ -127,3 +130,131 @@ func (d *Discharger) PublicKeyForLocation(loc string) (*bakery.PublicKey, error)
...
+			Message: fmt.Sprintf("invalid waitid %q", r.Form.Get("waitid")),
+		})
+		return
+	}
+	defer func() {
+		d.mu.Lock()
+		delete(d.waiting, r.Form.Get("waitid"))
+		d.mu.Unlock()
+	}()
+	err := <-discharge.c
```
rogpeppe: Perhaps have a (long) timeout here so that we don't have tests hanging forever if login FinishLogin is never called?

bakerytest/bakerytest.go
```diff
@@ -127,3 +130,131 @@ func (d *Discharger) PublicKeyForLocation(loc string) (*bakery.PublicKey, error)
...
+	if err != nil {
+		herr, ok := err.(*httpbakery.Error)
+		if !ok {
+			herr = &httpbakery.Error{
+				Code:    "forbidden",
```
rogpeppe: I think the code should be empty here. The idea of the code is that it's a set of known codes (defined in httpbakery/error.go). This isn't one of those.

rogpeppe: In fact, what about just using httpbakery.ErrorToResponse?

bakerytest/bakerytest.go
```diff
@@ -127,3 +130,131 @@ func (d *Discharger) PublicKeyForLocation(loc string) (*bakery.PublicKey, error)
...
+	}
+	m, err := d.Service.Discharge(
+		bakery.ThirdPartyCheckerFunc(
+			func(cavId, caveat string) ([]checkers.Caveat, error) {
+				return nil, nil
+			},
+		),
+		discharge.cavId,
+	)
+	if err != nil {
```
rogpeppe, line 261: I'd also use httpbakery.ErrorToResponse here too. I guess we _could_ just use httprequest's HandleErrors method and just return an error from this function.

bakerytest/bakerytest.go
```diff
@@ -127,3 +130,131 @@ func (d *Discharger) PublicKeyForLocation(loc string) (*bakery.PublicKey, error)
...
+		httprequest.WriteJSON(w, http.StatusForbidden, err)
+		return
+	}
+	httprequest.WriteJSON(
+		w,
+		http.StatusOK,
+		struct {
```
rogpeppe:
```
httpbakery.WaitResponse{
     Macaroon: m,
}
```

bakerytest/bakerytest.go
```diff
@@ -127,3 +130,131 @@ func (d *Discharger) PublicKeyForLocation(loc string) (*bakery.PublicKey, error)
...
+			Macaroon *macaroon.Macaroon
+		}{
+			Macaroon: m,
+		},
+	)
+}
+
+func (d *LoginDischarger) FinishLogin(w http.ResponseWriter, r *http.Request, err error) {
```
rogpeppe: doc comment please

bakerytest/bakerytest.go
```diff
@@ -127,3 +130,131 @@ func (d *Discharger) PublicKeyForLocation(loc string) (*bakery.PublicKey, error)
...
+	d.mu.Lock()
+	discharge, ok := d.waiting[r.Form.Get("waitid")]
+	d.mu.Unlock()
+	if !ok {
+		httprequest.WriteJSON(w, http.StatusBadRequest, httpbakery.Error{
+			Message: fmt.Sprintf("invalid waitid %q", r.Form.Get("waitid")),
+		})
+		return
+	}
+	discharge.c <- err
```
rogpeppe: Perhaps a timeout here? (not sure what best to do on timeout - panic?)

httpbakery/form/form.go
```diff
@@ -0,0 +1,144 @@
...
+	|   Client    |                            |
+	| Interaction |                            |
+	+-------------+                            |
+	      |                                    |
+	      | POST data to "form" URL            |
+	      |----------------------------------->|
+	      |                                    |
+	      |                Form login response |
+	      |<-----------------------------------|
+	      |                                    |
```
rogpeppe, line 45: This ASCII art is a marvel to behold.

httpbakery/form/form.go
```diff
@@ -0,0 +1,144 @@
...
+
+The schema is provided as a environschema.Fileds object. It is the
+client's responsibility to interpret the schema and present it to the
+user.
+*/
+
+// Objects implementing the Handler interface interpret schema and
```
rogpeppe: Rather than HandleForm, I wonder about GetForm, making it more clear what the role of the method is. Handlers are usually server side.

```
// FormGetter is implemented by clients to get the values for a form schema.
// For example, an implementation might prompt the user for the value of
// each field in the schema. The value returned from GetForm must be
// compatible with the provided schema.
type FormGetter interface {
     GetForm(schema environschema.Fields) (map[string]interface{}, error)
}
```

?

httpbakery/form/form.go
```diff
@@ -0,0 +1,144 @@
...
+// sending to the server. Any errors returned by h.Handle will not have
+// their cause masked.
+func SetUpAuth(c *httpbakery.Client, h Handler) {
+	v := webPageVisitor{
+		client: &httprequest.Client{
+			Doer: c,
+		},
+		handler: h,
+	}
+	c.VisitWebPage = v.visitWebPage
```
rogpeppe, line 75: or inline the webPageVisitor creation?

httpbakery/form/form.go
```diff
@@ -0,0 +1,144 @@
...
+// loginMethods contains the response expected from the login URL. It
+// only checks for the "form" method as that is the only one that can be
+// handled.
+type loginMethods struct {
+	Form string `json:"form"`
+}
+
+// schemaResponse contains the message expected in response to the schema
+// request.
+type schemaResponse struct {
```
rogpeppe: I wonder if this should be exported so that the server can use it.

httpbakery/form/form.go
```diff
@@ -0,0 +1,144 @@
...
+	Schema environschema.Fields `json:"schema"`
+}
+
+type loginRequest struct {
+	httprequest.Route `httprequest:"POST"`
+	Login             login `httprequest:",body"`
+}
+
+type login struct {
```
rogpeppe: Similarly here.

httpbakery/form/form.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	if lm.Form == "" {
+		return errgo.Newf("form login not supported")
+	}
+	var s schemaResponse
+	if err := v.client.Get(lm.Form, &s); err != nil {
+		return errgo.Notef(err, "cannot get schema")
+	}
+	if len(s.Schema) == 0 {
+		return errgo.Newf("invalid schema")
```
rogpeppe: "invalid schema: no fields found"

?

httpbakery/form/form.go
```diff
@@ -0,0 +1,144 @@
...
+	Form string `json:"form"`
+}
+
+// schemaResponse contains the message expected in response to the schema
+// request.
+type schemaResponse struct {
+	Schema environschema.Fields `json:"schema"`
+}
+
+type loginRequest struct {
```
rogpeppe: I know it was a hassle but I like the way this looks, thanks.

bakerytest/bakerytest.go
```diff
@@ -127,3 +130,131 @@ func (d *Discharger) PublicKeyForLocation(loc string) (*bakery.PublicKey, error)
...
+// loginHandler to handle the login interraction. The login is completed
+// by calling FinishLogin.
+//
+// If locator is non-nil, it will be used to find public keys
+// for any third party caveats returned by the checker.
+//
+// Calling this function has the side-effect of setting
+// InsecureSkipVerify in http.DefaultTransport.TLSClientConfig
+// until all the dischargers are closed.
+func NewLoginDischarger(locator bakery.PublicKeyLocator, loginHandler http.Handler) *LoginDischarger {
```
rogpeppe: Given the terminology we use elsewhere in the bakery, I wonder whether this should be "NewInteractiveDischarger", and loginHandler could be visitHandler.

I also wonder if, instead of implementing http.Handler, loginHandler could be:

```
func(http.ResponseWriter, *http.Request) error
```

and we could call FinishLogin automatically when it returns (and unexport FinishLogin).

httpbakery/form/form_test.go
```diff
@@ -0,0 +1,219 @@
...
+		Scheme: ":",
+	}
+	err := client.VisitWebPage(&u)
+	c.Assert(err, gc.ErrorMatches, "cannot create request: parse :://: missing protocol scheme")
+}
+
+type formDischarger struct {
+	discharger *bakerytest.LoginDischarger
+
+	ignoreAccept    bool
```
rogpeppe, line 119: How about putting all these fields in a struct type so that we
can copy the whole lot easily in the test? Makes the tests
slightly more verbose, but probably worth it.

bakerytest/bakerytest.go
mhilton: No, we don't necessarily want to finish the first time this handler is called, especially in the case this was written for.

rogpeppe: Ah, having read a bit further below, I now see this isn't possible. Thinking more...

httpbakery/form/form_test.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+	if d.loginError {
+		httprequest.WriteJSON(w, http.StatusInternalServerError, testError)
+		d.discharger.FinishLogin(w, r, testError)
+		return
+	}
+	methods := map[string]string{}
+	if !d.formUnsupported {
+		r.ParseForm()
+		methods["form"] = d.discharger.Location() + "/form?waitid=" + r.Form.Get("waitid")
```
rogpeppe: Hmm, this seems like unwarranted chumminess with the LoginDischarger (it doesn't document waitid anywhere).

I wonder if LoginDischarger could provide a Context method
e.g.

```
// Context returns the discharger request context from the
// given request. It returns an empty string if no context is found.
Context(req *http.Request) string
```

Then FinishLogin could be:

```
// FinishLogin completes a login request. The context should
// have been obtained from the original call to the login handler.
// It panics if context is not valid.
FinishLogin(w http.ResponseWriter, context string, err error)
```

Then we wouldn't need to know anything about waitid here.

httpbakery/form/form_test.go
```diff
@@ -0,0 +1,219 @@
...
+	if err != nil {
+		d.errorf(w, r, "bad login request: %s", err)
+		return
+	}
+	d.discharger.FinishLogin(w, r, nil)
+}
+
+func (d *formDischarger) errorf(w http.ResponseWriter, r *http.Request, s string, p ...interface{}) {
+	err := &httpbakery.Error{
+		Code:    "bad request",
```
rogpeppe: httpbakery.ErrBadRequest

rogpeppe:
This is really nice stuff, thanks. I have a bunch of comments and suggestions, but nothing major.

bakerytest/bakerytest.go
mhilton: done

bakerytest/bakerytest.go
mhilton: done

bakerytest/bakerytest.go
mhilton, line 186: done

bakerytest/bakerytest.go
mhilton: why not indeed, done

bakerytest/bakerytest.go
mhilton: done

bakerytest/bakerytest.go
mhilton: done

bakerytest/bakerytest.go
mhilton, line 261: done

bakerytest/bakerytest.go
mhilton: done

bakerytest/bakerytest.go
mhilton: done

bakerytest/bakerytest.go
mhilton: done, the timeout is zero (there is something wrong if it doesn't send immediately)

httpbakery/form/form.go
mhilton, line 45: thanks

httpbakery/form/form.go
mhilton: I went with Processor in the end (I didn't like Getter)

httpbakery/form/form.go
mhilton, line 75: I prefer it this way, the code feels too clunky the other way.

httpbakery/form/form.go
mhilton: done

httpbakery/form/form.go
mhilton: done

httpbakery/form/form.go
mhilton: done

httpbakery/form/form_test.go
mhilton, line 119: done

httpbakery/form/form_test.go
mhilton: I've done this another way, but now the handler doesn't know about anything internal to LoginDischarger

httpbakery/form/form_test.go
mhilton: done

bakerytest/bakerytest.go
mhilton: done

bakerytest/bakerytest.go
```diff
@@ -127,3 +131,174 @@ func (d *Discharger) PublicKeyForLocation(loc string) (*bakery.PublicKey, error)
...
+	Mux *http.ServeMux
+
+	// mu protects the following fields.
+	mu      sync.Mutex
+	waiting map[string]discharge
+	id      int
+}
+
+// NewInteractiveDischarger returns a new InteractiveDischarger. The
+// InteractiveDischarger will server the following endpoints by default:
```
rogpeppe: s/server/serve/

bakerytest/bakerytest.go
```diff
@@ -127,3 +131,174 @@ func (d *Discharger) PublicKeyForLocation(loc string) (*bakery.PublicKey, error)
...
+	}
+	select {
+	case discharge.c <- err:
+	default:
+		panic("cannot finish interaction " + r.Form.Get("waitid"))
+	}
+	return
+}
+
+// URL generates a URL addressed path that will be handled correctly by
```
rogpeppe: "URL addressed path" reads oddly to me.

// URL returns a URL addressed to the given path in the discharger
// that contains any discharger context information found
// in the given request. Use this to generate intermediate URLs
// before calling FinishInteraction.

?

httpbakery/form/form.go
```diff
@@ -0,0 +1,151 @@
...
+	      |                Form login response |
+	      |<-----------------------------------|
+	      |                                    |
+
+The schema is provided as a environschema.Fileds object. It is the
+client's responsibility to interpret the schema and present it to the
+user.
+*/
+
+// Objects implementing the FormProcessor interface interpret schema and
```
rogpeppe: s/FormProcessor/Processor/

I still think that "Get" (which is essentially the contract of this function) is
more straightforward than the uncertainly-worded "Process".

Hmm, thinking further into form analogies, how about Filler ?

```
// Filler represents an object that can fill out a form. The given schema
// represents the form template. The returned value should be compatible
// with this.
```

We don't really need to talk about user interaction here - that might or might
not happen, depending on the client.

httpbakery/form/form.go
```diff
@@ -0,0 +1,151 @@
...
+// interact with the user. When the interaction is complete the returned
+// value should conform to the provided schema.
+type Processor interface {
+	ProcessForm(schema environschema.Fields) (interface{}, error)
```
rogpeppe: Shouldn't this return map[string] interface{} ?

httpbakery/form/form.go
mhilton: I don't like "get" because that to me seems synonymous with collect. "Filler" I like.

httpbakery/form/form.go
```diff
@@ -0,0 +1,151 @@
...
+// SetUpAuth configures form authentication on c. The VisitWebPage field
+// in c will be set to a function that will attempt form based
+// authentication using h to perform the interaction with the user.
+//
+// The VisitWebPage function downloads the schema from the specified
+// server and calls h.Handle. The map returned by h.Handle should match
+// the schema specified, but VisitWebPage will not verify this before
+// sending to the server. Any errors returned by h.Handle will not have
+// their cause masked.
+func SetUpAuth(c *httpbakery.Client, p Processor) {
```
rogpeppe: Thinking about this a bit more, I really think it's important (in the near future, not necessarily in this PR) to enable fallback to other methods. For example, I might want to use form-based login in preference, but fall back to web page interaction when necessary.

One possibility here is to chain the VisitWebPage functions, so that
the VisitWebPage value set up by SetUpAuth will fall back to calling the
original VisitWebPage value if it doesn't find the requisite method.

httpbakery/form/form.go
mhilton: I don't think it necessarily has to, although that is the most likely thing to return. There may be reasons to use something else.

httpbakery/form/form.go
```diff
@@ -0,0 +1,151 @@
...
+
+// LoginRequest is a request to perform a login using the provided details.
+type LoginRequest struct {
+	httprequest.Route `httprequest:"POST"`
+	Login             LoginWrapper `httprequest:",body"`
+}
+
+// Login holds the body of a login request.
+type LoginWrapper struct {
+	Login interface{} `json:"login"`
```
rogpeppe: map[string] interface{}

?

httpbakery/form/form.go
```diff
@@ -0,0 +1,151 @@
...
+	if err != nil {
+		return errgo.Notef(err, "cannot create request")
+	}
+	req.Header.Set("Accept", "application/json")
+	var lm loginMethods
+	if err := v.client.Do(req, nil, &lm); err != nil {
+		return errgo.Notef(err, "cannot get login methods")
+	}
+	if lm.Form == "" {
+		return errgo.Newf("form login not supported")
```
rogpeppe, line 129: potentially chain to next visitWebPage function here.

httpbakery/form/form.go
mhilton: that seems reasonable, I'll do it in another PR, the mechanics need some thought.

rogpeppe:
Nice updates. LGTM modulo a few more suggestions, thanks!

httpbakery/form/form.go
mhilton: This one specifically I don't want to be a map, if it is an interface I can do creative things on servers with real types.

httpbakery/form/form.go
urosj: Hm. If we go this route in the future, we need to make it "predictable" and well defined. There are environments, where web is not an option.

httpbakery/form/form.go
mhilton: done

httpbakery/form/form.go
mhilton: done

httpbakery/form/form.go
mhilton: done

urosj:
:+1: looking awesome

bakerytest/bakerytest.go
mhilton: done

bakerytest/bakerytest.go
mhilton: done

PR #64, opened 2015-09-22
use paths set in macaroon discharges without change

mhilton:
If a discharge required error set a macaroon path then the cookie path used will match this path exactly.

### DIFF
--- a/httpbakery/client.go
+++ b/httpbakery/client.go
@@ -261,18 +261,25 @@ func (c *Client) doWithBody(req *http.Request, body io.ReadSeeker) (*http.Respon
 	if err != nil {
 		return nil, errgo.Mask(err, errgo.Any)
 	}
-	cookieURL := req.URL
+	var cookiePath string
 	if path := resp.Info.MacaroonPath; path != "" {
 		relURL, err := parseURLPath(path)
 		if err != nil {
 			logger.Warningf("ignoring invalid path in discharge-required response: %v", err)
 		} else {
-			cookieURL = req.URL.ResolveReference(relURL)
+			cookiePath = req.URL.ResolveReference(relURL).Path
 		}
 	}
-	if err := SetCookie(c.Jar, cookieURL, macaroons); err != nil {
-		return nil, errgo.Notef(err, "cannot set cookie")
+	if cookiePath != "/" {
+		cookiePath = strings.TrimSuffix(cookiePath, "/")
 	}
+	cookie, err := NewCookie(macaroons)
+	if err != nil {
+		return nil, errgo.Notef(err, "cannot make cookie")
+	}
+	cookie.Path = cookiePath
+	c.Jar.SetCookies(req.URL, []*http.Cookie{cookie})
+
 	if err := c.setRequestBody(req, body); err != nil {
 		return nil, errgo.Mask(err)
 	}
--- a/httpbakery/client_test.go
+++ b/httpbakery/client_test.go
@@ -256,7 +256,6 @@ func (s *ClientSuite) TestMacaroonCookiePath(c *gc.C) {
 	assertCookieCount := func(path string, n int) {
 		u, err := url.Parse(ts.URL + path)
 		c.Assert(err, gc.IsNil)
-		c.Logf("client jar %p", client.Jar)
 		c.Assert(client.Jar.Cookies(u), gc.HasLen, n)
 	}
 	cookiePath = ""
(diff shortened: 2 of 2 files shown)


httpbakery/client.go
```diff
@@ -338,10 +345,18 @@ func NewCookie(ms macaroon.Slice) (*http.Cookie, error) {
 // contain a single primary macaroon in its first element, and any
 // discharges after that.
 func SetCookie(jar http.CookieJar, url *url.URL, ms macaroon.Slice) error {
+	return SetCookieWithPath(jar, url, "", ms)
+}
+
+// SetCookieWithPath is like SetCookie except the cookie path is
+// explicitely specified. Calling SetCookieWithPath with a path of "" is
```
fabricematrat: /s/explicitely/explicitly

httpbakery/client.go
```diff
@@ -261,18 +261,25 @@ func (c *Client) doWithBody(req *http.Request, body io.ReadSeeker) (*http.Respon
...
+			cookieURL := req.URL.ResolveReference(relURL)
+			cookiePath = cookieURL.Path
 		}
 	}
-	if err := SetCookie(c.Jar, cookieURL, macaroons); err != nil {
+	if len(cookiePath) > 1 && strings.HasSuffix(cookiePath, "/") {
+		cookiePath = strings.TrimSuffix(cookiePath, "/")
+	}
+	logger.Debugf("cookiePath: %q", cookiePath)
+	if err := SetCookieWithPath(c.Jar, cookieURL, cookiePath, macaroons); err != nil {
```
rogpeppe: I'm not entirely sure that SetCookieWithPath is justified. I think I'd just use NewCookie, set the path and use c.Jar.SetCookies.

That way we perhaps won't feel tempted to add SetCookieWithPathAndName and SetCookieWithName too.

jrwren:
:+1:

httpbakery/client.go
```diff
@@ -261,18 +261,25 @@ func (c *Client) doWithBody(req *http.Request, body io.ReadSeeker) (*http.Respon
...
 	}
+	var cookiePath string
 	cookieURL := req.URL
 	if path := resp.Info.MacaroonPath; path != "" {
 		relURL, err := parseURLPath(path)
 		if err != nil {
 			logger.Warningf("ignoring invalid path in discharge-required response: %v", err)
 		} else {
-			cookieURL = req.URL.ResolveReference(relURL)
+			cookieURL := req.URL.ResolveReference(relURL)
```
mhilton: The change in this line is wrong, we should be setting the existing cookieURL.

httpbakery/client_test.go
```diff
@@ -278,6 +278,16 @@ func (s *ClientSuite) TestMacaroonCookiePath(c *gc.C) {
 	assertCookieCount("/foo/bar/", 1)
 	assertCookieCount("/foo/bar/baz", 1)
 
+	cookiePath = "/foo"
```
rogpeppe, line 280: I wonder if we should have a test here with a relative path (e.g. ../something)

httpbakery/client.go
```diff
@@ -338,10 +345,18 @@ func NewCookie(ms macaroon.Slice) (*http.Cookie, error) {
 // contain a single primary macaroon in its first element, and any
 // discharges after that.
 func SetCookie(jar http.CookieJar, url *url.URL, ms macaroon.Slice) error {
+	return SetCookieWithPath(jar, url, "", ms)
+}
+
+// SetCookieWithPath is like SetCookie except the cookie path is
```
rogpeppe: I'm not sure that this pulls its weight. I'd just document in SetCookie that if you want to set the cookie for a particular path or with a given name, then use NewCookie, set the relevant fields and then use CookieJar.SetCookies.

httpbakery/client.go
```diff
@@ -261,18 +261,25 @@ func (c *Client) doWithBody(req *http.Request, body io.ReadSeeker) (*http.Respon
...
 		if err != nil {
 			logger.Warningf("ignoring invalid path in discharge-required response: %v", err)
 		} else {
-			cookieURL = req.URL.ResolveReference(relURL)
+			cookieURL := req.URL.ResolveReference(relURL)
+			cookiePath = cookieURL.Path
 		}
 	}
-	if err := SetCookie(c.Jar, cookieURL, macaroons); err != nil {
+	if len(cookiePath) > 1 && strings.HasSuffix(cookiePath, "/") {
```
rogpeppe: if think:

   cookiePath = strings.TrimSuffix(cookiePath, "/")

should be sufficient here.

httpbakery/client.go
rogpeppe: Given that you're setting the cookie path here, why not just use req.URL ?

httpbakery/client.go
fabricematrat: @rogpeppe  what if cookiePath is "/" ?

rogpeppe:
Looks reasonable, thanks, but I have a few suggestions you might want to consider.

httpbakery/client_test.go
mhilton, line 280: Because of the way the test works it can't be anything outside of the request path otherwise doRequest fails. but I've done "../bar".

httpbakery/client.go
mhilton: @rogpeppe fabrice is correct here. that's why.

fabricematrat:
LGTM after not using SetCookie like @rogpeppe  suggested.

httpbakery/client.go
```diff
@@ -261,18 +261,25 @@ func (c *Client) doWithBody(req *http.Request, body io.ReadSeeker) (*http.Respon
...
 		if err != nil {
 			logger.Warningf("ignoring invalid path in discharge-required response: %v", err)
 		} else {
-			cookieURL = req.URL.ResolveReference(relURL)
+			cookiePath = req.URL.ResolveReference(relURL).Path
 		}
 	}
-	if err := SetCookie(c.Jar, cookieURL, macaroons); err != nil {
-		return nil, errgo.Notef(err, "cannot set cookie")
+	if len(cookiePath) > 1 {
```
rogpeppe: if cookiePath != "/" {

might be a more direct way to phrase the intention here?

rogpeppe:
LGTM thanks

PR #66, opened 2015-09-23
httpbakery: allow custom error responses

rogpeppe:
We have a need to deal with non-httpbakery-like error messages
so we allow that freedom in httpbakery.Client.Do.

### DIFF
--- a/httpbakery/client.go
+++ b/httpbakery/client.go
@@ -213,13 +213,33 @@ func relativeURL(base, new string) (*url.URL, error) {
 //
 // Do may add headers to req.Header.
 func (c *Client) DoWithBody(req *http.Request, body io.ReadSeeker) (*http.Response, error) {
+	return c.DoWithBodyAndCustomError(req, body, nil)
+}
+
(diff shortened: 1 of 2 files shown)


makyo:
:+1:

httpbakery/client.go
```diff
@@ -213,13 +213,33 @@ func relativeURL(base, new string) (*url.URL, error) {
...
+// a non-nil Error if the response holds an error that
+// can be converted to an Error value. The getError call must leave the
+// response body unchanged otherwise.
+//
+// If getError is nil, DefaultGetError will be used.
+//
+// This method can be useful when dealing with APIs that
+// return their errors in a format incompatible with Error, but the
+// need for it should be avoided when creating new APIs,
+// as it makes the endpoints not amenable to generic tools.
```
mhilton: "less" as opposed to "not"?

mhilton:
:+1:

httpbakery/client.go

PR #67, opened 2015-09-24
simple CLI form client

mhilton:
Simple client for filling out form logins on the command line.

### DIFF
--- a/httpbakery/form/client.go
+++ b/httpbakery/form/client.go
@@ -1,0 +1,201 @@
+package form
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"sort"
+	"strings"
+
+	"github.com/juju/schema"
+	"golang.org/x/crypto/ssh/terminal"
+	"gopkg.in/errgo.v1"
+	"gopkg.in/juju/environschema.v1"
+)
+
+// PromptingFiller implements a Filler by prompting for each field in
+// turn.
+type PromptingFiller struct {
+	// The Prompter to use to get the responses if this is nil then
+	// DefaultPrompter will be used.
+	Prompter Prompter
+}
+
+// Fill processes fields by first sorting them and then prompting for the
+// value of each one in turn.
+//
+// The fields are sorted by first by group name. Those in the same group
+// are sorted so that secret fields come after non-secret ones, finally
+// the fields are sorted by description.
+//
+// Each field will be prompted for, the returned value will then be
+// validated against the type for the field. If the returned value does
+// not validate correctly it will be prompted again twice before giving
+// up.
+//
+// If a field contains environment variables these will be checked in
+// order. The first one that is set will be used as the value of default
+// when Prompt is called.
+func (f PromptingFiller) Fill(fields environschema.Fields) (map[string]interface{}, error) {
+	fs := make(fieldSlice, len(fields))
+	var i int
+	for k, v := range fields {
+		fs[i] = field{
+			name:  k,
+			attrs: v,
+		}
+		i++
+	}
+	sort.Sort(fs)
+	form := make(map[string]interface{}, len(fields))
+	for _, field := range fs {
+		var err error
+		form[field.name], err = f.prompt(field.attrs)
+		if err != nil {
+			return nil, errgo.Notef(err, "cannot complete form")
+		}
+	}
+	return form, nil
+}
+
+type field struct {
+	name  string
+	attrs environschema.Attr
+}
+
+type fieldSlice []field
+
+func (s fieldSlice) Len() int {
+	return len(s)
+}
+
+func (s fieldSlice) Less(i, j int) bool {
+	a1 := s[i].attrs
+	a2 := s[j].attrs
+	if a1.Group != a2.Group {
+		return a1.Group < a2.Group
+	}
+	if a1.Secret != a2.Secret {
+		return a2.Secret
+	}
+	return a1.Description < a2.Description
+}
+
+func (s fieldSlice) Swap(i, j int) {
+	s[i], s[j] = s[j], s[i]
+}
+
+func (f *PromptingFiller) prompt(attr environschema.Attr) (interface{}, error) {
+	prompter := f.Prompter
+	if prompter == nil {
+		prompter = DefaultPrompter
+	}
+	def := f.getDefault(attr)
+	for i := 0; i < 3; i++ {
+		val, err := prompter.Prompt(attr.Description, def, attr.Secret)
+		if err != nil {
+			return nil, errgo.Notef(err, "cannot get input")
+		}
+		switch attr.Type {
+		case environschema.Tbool:
+			b, err := schema.Bool().Coerce(val, nil)
+			if err == nil {
+				return b, nil
+			}
+		case environschema.Tint:
+			i, err := schema.Int().Coerce(val, nil)
+			if err == nil {
+				return i, nil
+			}
+		default:
+			return val, nil
+		}
+	}
+	return nil, errgo.New("too many invalid inputs")
+}
+
+func (f *PromptingFiller) getDefault(attr environschema.Attr) string {
+	if attr.EnvVar != "" {
+		if env := os.Getenv(attr.EnvVar); env != "" {
+			return env
+		}
+	}
+	for _, envVar := range attr.EnvVars {
+		if env := os.Getenv(envVar); env != "" {
+			return env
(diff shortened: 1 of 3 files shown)


httpbakery/form/client.go
```diff
@@ -0,0 +1,201 @@
...
+}
+
+// Fill processes fields by first sorting them and then prompting for the
+// value of each one in turn.
+//
+// The fields are sorted by first by group name. Those in the same group
+// are sorted so that secret fields come after non-secret ones, finally
+// the fields are sorted by description.
+//
+// Each field will be prompted for, the returned value will then be
```
rogpeppe, line 31: s/the/then the/ ?

httpbakery/form/client.go
```diff
@@ -0,0 +1,201 @@
...
+// validated against the type for the field. If the returned value does
+// not validate correctly it will be prompted again twice before giving
+// up.
+//
+// If a field contains environment variables these will be checked in
+// order. The first one that is set will be used as the value of default
+// when Prompt is called.
+func (f PromptingFiller) Fill(fields environschema.Fields) (map[string]interface{}, error) {
+	fs := make(fieldSlice, len(fields))
```
rogpeppe, line 40: My preferred idiom for this kind of thing is:

```
fs := make(fieldSlice, 0, len(fields))
```

then use append in the loop. Then you don't have to maintain the index.

httpbakery/form/client.go
```diff
@@ -0,0 +1,201 @@
...
+	s[i], s[j] = s[j], s[i]
+}
+
+func (f *PromptingFiller) prompt(attr environschema.Attr) (interface{}, error) {
+	prompter := f.Prompter
+	if prompter == nil {
+		prompter = DefaultPrompter
+	}
+	def := f.getDefault(attr)
+	for i := 0; i < 3; i++ {
```
rogpeppe, line 94: Perhaps make this count a field in the PromptingFiller ?

httpbakery/form/client.go
```diff
@@ -0,0 +1,201 @@
...
+		return a2.Secret
+	}
+	return a1.Description < a2.Description
+}
+
+func (s fieldSlice) Swap(i, j int) {
+	s[i], s[j] = s[j], s[i]
+}
+
+func (f *PromptingFiller) prompt(attr environschema.Attr) (interface{}, error) {
```
rogpeppe, line 88: It's probably best to be consistent in whether PromptingFiller is passed by value or reference. (Fill has it passed by value). I don't care which.

httpbakery/form/client.go
```diff
@@ -0,0 +1,201 @@
...
+			b, err := schema.Bool().Coerce(val, nil)
+			if err == nil {
+				return b, nil
+			}
+		case environschema.Tint:
+			i, err := schema.Int().Coerce(val, nil)
+			if err == nil {
+				return i, nil
+			}
+		default:
```
rogpeppe, line 110: I think this should probably do a bit better if an unknown type is encountered (for example Tattrs). I'd suggest testing for environschema.Tstring explicitly and returning an error if an unknown type is encountered.

httpbakery/form/client.go
```diff
@@ -0,0 +1,201 @@
...
+	"sort"
+	"strings"
+
+	"github.com/juju/schema"
+	"golang.org/x/crypto/ssh/terminal"
+	"gopkg.in/errgo.v1"
+	"gopkg.in/juju/environschema.v1"
+)
+
+// PromptingFiller implements a Filler by prompting for each field in
```
urosj, line 16: It would probably be nice to start with a general description what we aim to do with this package.

httpbakery/form/client.go
```diff
@@ -0,0 +1,201 @@
...
+			return env
+		}
+	}
+	return ""
+}
+
+// A Prompter is used by a PromptingFiller to get the values of the
+// fields.
+type Prompter interface {
+	Prompt(description, def string, secret bool) (string, error)
```
rogpeppe, line 134: How about:

```
 // Prompt prompts for the value of the given named attribute.
 // It is always acceptable for an implementation to return a string,
 // which will be coerced to the appropriate type if possible.
 Prompt(name string, environschema.Attr) (interface{}, error)
```

Then an implementation can use as much or as little information
of the attribute as it likes to determine how to prompt, and we don't
need to worry about new fields being added to environschema.Attr
that might need to be passed to Prompt and thus break the Prompter interface.

httpbakery/form/client.go
```diff
@@ -0,0 +1,201 @@
...
+}
+
+// DefaultPrompter is the default Prompter used by a PromptingFiller when
+// Prompter has not been set.
+var DefaultPrompter Prompter = IOPrompter{
+	In:  os.Stdin,
+	Out: os.Stderr,
+}
+
+// IOPrompter is a Prompter based around an io.Reader and io.Writer.
```
rogpeppe, line 144: // It uses Out to print prompts to and In to read responses from.

?

httpbakery/form/client.go
```diff
@@ -0,0 +1,201 @@
...
+	prompt := description
+	if def != "" {
+		def1 := def
+		if secret {
+			def1 = strings.Repeat("*", len(def))
+		}
+		prompt = fmt.Sprintf("%s (%s)", description, def1)
+	}
+	prompt += ": "
+	_, err := fmt.Fprintf(p.Out, "%s", prompt)
```
rogpeppe, line 164: or fmt.Fprint(p.Out, prompt)

httpbakery/form/interactive_test.go
```diff
@@ -0,0 +1,83 @@
...
+	about:       "prompt with default",
+	message:     `Please enter "pass" at the following prompt (there should be no echo)`,
+	description: "test",
+	def:         "fail",
+	secret:      true,
+	expect:      "pass",
+}}
+
+func (s *interactiveSuite) TestIOPrompter(c *gc.C) {
+	f, err := os.OpenFile("/dev/tty", os.O_RDWR, 0666)
```
rogpeppe, line 69: An alternative to this might be to have a separate binary that you can run directly with go run. Then we could let the test run on Windows too.

rogpeppe:
This is really nice, thanks. Along with the above suggestions, can I also suggest that this gets re-proposed against the environschema repo, perhaps named as package environschemaform ?

httpbakery/form/client.go
mhilton, line 88: done

httpbakery/form/client.go
mhilton, line 94: done

httpbakery/form/client.go
mhilton, line 164: I have kept this the same, but incorporated the ": " into the format string.

httpbakery/form/client.go
```diff
@@ -0,0 +1,201 @@
...
+	if f, ok := r.(*os.File); ok && secret && terminal.IsTerminal(int(f.Fd())) {
+		defer w.Write([]byte{'\n'})
+		return terminal.ReadPassword(int(f.Fd()))
+	}
+	var input []byte
+	for {
+		var buf [1]byte
+		n, err := r.Read(buf[:])
+		if n == 1 {
+			if buf[0] == '\n' {
```
urosj, line 188: On windows, don't we have \r\n? Assuming here, that we need to support clients from Windows that is :)

httpbakery/form/client.go
mhilton, line 134: done

httpbakery/form/client.go
mhilton, line 110: done

urosj:
Looks good, however, please add TODO on marked places for future windows compatibility.

httpbakery/form/client.go
mhilton, line 144: done

httpbakery/form/interactive_test.go
mhilton, line 69: done

httpbakery/form/client.go
mhilton, line 40: done

httpbakery/form/client.go
mhilton, line 31: done

httpbakery/form/client.go
mhilton, line 188: I think we should, so done.

PR #68, opened 2015-09-24
bakery: return an error if CheckAny is called with no macaroons

rogpeppe:
CheckAny would erroneously succeed when provided with no macaroons.
In most situations this wasn't a problem in practice because most clients
use httpbakery.CheckRequest which does its own check for this condition,
but it's still a potential ravening security hole.

Also add a set of tests for CheckAny because there were none before.


bakery/service.go
```diff
@@ -144,6 +144,11 @@ func (svc *Service) Check(ms macaroon.Slice, checker FirstPartyChecker) error {
 //
 // It returns any attributes declared in the successfully validated request.
```
jrwren: Document that it returns an error if mss is empty.

mhilton:
:+1: LGTM

jrwren:
:+1:

PR #69, opened 2015-09-27
Allow customizing discharge acquisition.

cmars:
Added a DischargeAcquirer field to httpbakery.Client, which is an interface that is used to obtain third-party discharge macaroons. If not set, it falls back to the HTTP implementation provided by the Client itself.


cmars:
@rogpeppe PTAL, thanks!

rogpeppe:
Nice suggestion. LGTM but I'd like to see a test.

cmars:
Added a test.

PR #70, opened 2015-09-29
httpbakery: return generic error from getError

rogpeppe:
The new DoWithBodyAndCustomError turned out to be awkward to
use in practice. This makes things a bit easier to use - it's possible
to return any kind of error from the getError function and we explicitly
document that the discharge-required error is the only kind of error
that we really care about.

Note that technically this is a backwardly incompatible change, but
the API has only existed for a few days so there are very unlikely
to be clients already reliant on it.


mhilton:
:+1: LGTM

frankban:
:+1:

PR #71, opened 2015-09-29
keep trailing / on cookie path

mhilton:
This will let the macaroon be created with the same path as the cookie
set by the wait endpoint. Reducing the number of cookies sent.


httpbakery/client.go
```diff
@@ -286,9 +286,6 @@ func (c *Client) doWithBody(req *http.Request, body io.ReadSeeker, getError func
 			cookiePath = req.URL.ResolveReference(relURL).Path
 		}
 	}
-	if cookiePath != "/" {
```
rogpeppe, line 289: Cool. I did wonder why this was there.

rogpeppe:
LGTM, thanks.

PR #72, opened 2015-09-30
Add fallback option if form login is not supported.

mhilton:
Also enable creation of a VisitWebPage function without an existing
httpbakery.Client for more flexible configuration.

### DIFF
--- a/httpbakery/form/form.go
+++ b/httpbakery/form/form.go
@@ -57,28 +57,44 @@ type Filler interface {
 }
 
 // SetUpAuth configures form authentication on c. The VisitWebPage field
-// in c will be set to a function that will attempt form based
-// authentication using h to perform the interaction with the user.
-//
-// The VisitWebPage function downloads the schema from the specified
-// server and calls h.Handle. The map returned by h.Handle should match
-// the schema specified, but VisitWebPage will not verify this before
-// sending to the server. Any errors returned by h.Handle will not have
-// their cause masked.
+// in c will be set to a function that will attempt form-based
+// authentication using f to perform the interaction with the user and
+// fall back to using the current value of VisitWebPage if form-based
+// authentication is not supported.
 func SetUpAuth(c *httpbakery.Client, f Filler) {
-	v := webPageVisitor{
-		client: &httprequest.Client{
+	c.VisitWebPage = VisitWebPage(
+		&httprequest.Client{
 			Doer: c,
 		},
-		filler: f,
+		f,
+		c.VisitWebPage,
+	)
+}
+
+// VisitWebPage creates a function suitable for use with
+// httpbakery.Client.VisitWebPage. The new function downloads the schema
+// from the specified server and calls f.Fill. The map returned by f.Fill
+// should match the schema specified, but this is not verified before
+// sending to the server. Any errors returned by f.Fill or fallback will
+// not have their cause masked.
+//
+// If the new function detects that form login is not supported by the
+// server and fallback is not nil then fallback will be called to perform
+// the visit.
+func VisitWebPage(c *httprequest.Client, f Filler, fallback func(u *url.URL) error) func(u *url.URL) error {
+	v := webPageVisitor{
+		client:   c,
+		filler:   f,
+		fallback: fallback,
 	}
-	c.VisitWebPage = v.visitWebPage
+	return v.visitWebPage
 }
 
 // webPageVisitor contains the state required by visitWebPage.
 type webPageVisitor struct {
-	client *httprequest.Client
-	filler Filler
+	client   *httprequest.Client
+	filler   Filler
+	fallback func(u *url.URL) error
 }
 
 // loginMethods contains the response expected from the login URL. It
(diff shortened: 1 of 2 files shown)


httpbakery/form/form.go
```diff
@@ -58,27 +58,44 @@ type Filler interface {
...
+		f,
+		c.VisitWebPage,
+	)
+}
+
+// VisitWebPage creates a function suitable for use with
+// httpbakery.Client.VisitWebPage. The new function downloads the schema
+// from the specified server and calls f.Fill. The map returned by f.Fill
+// should match the schema specified, but this is not verified before
+// sending to the server. Any errors returned by f.Fill will not have
```
urosj: Server should be able to handle unmatched schemas anyway. But it would also be nice to minimize number of false requests if possible. Suggestion for next PR maybe.

httpbakery/form/form.go
```diff
@@ -58,27 +58,44 @@ type Filler interface {
...
+// their cause masked.
+//
+// If the new function detects that form login is not supported by the
+// server and fb is not nill then fb will be called to perform the visit.
```
urosj: s/nill/nil

httpbakery/form/form.go
```diff
@@ -58,27 +58,44 @@ type Filler interface {
 
 // SetUpAuth configures form authentication on c. The VisitWebPage field
 // in c will be set to a function that will attempt form based
-// authentication using h to perform the interaction with the user.
+// authentication using f to perform the interaction with the user.
```
rogpeppe: There seems to be some overlap between these two sentences.

Perhaps just:

```
 // SetUpAuth configures form authentication on c. The VisitWebPage field
 // in c will be set to a function that will attempt form-based
 // authentication using f to perform the interaction with the user
 // and fall back to using the current value of VisitWebPage
 // if form-based authentication is not supported.
```

?

httpbakery/form/form.go
rogpeppe: I think it's reasonable for the Filler to have responsibility for producing well-formed data - it has all the information necessary to do so.

httpbakery/form/form.go
```diff
@@ -58,27 +58,44 @@ type Filler interface {
...
+// VisitWebPage creates a function suitable for use with
+// httpbakery.Client.VisitWebPage. The new function downloads the schema
+// from the specified server and calls f.Fill. The map returned by f.Fill
+// should match the schema specified, but this is not verified before
+// sending to the server. Any errors returned by f.Fill will not have
+// their cause masked.
+//
+// If the new function detects that form login is not supported by the
+// server and fb is not nill then fb will be called to perform the visit.
+func VisitWebPage(c *httprequest.Client, f Filler, fb func(u *url.URL) error) func(u *url.URL) error {
```
rogpeppe: I think I'd probably call it fallback not fb to emphasise the relationship with webPageVisitor.fallback. I don't mind much though.

mhilton: I was thinking along those lines too. As the type gives little information the name probably should.

httpbakery/form/form.go
rogpeppe: // Any errors returned from f.Fill or fb will not have their cause masked.

would be slightly more accurate, I think.

mhilton: done (the last one)

httpbakery/form/form.go
mhilton: done

httpbakery/form/form.go
```diff
@@ -123,9 +140,21 @@ func (v webPageVisitor) visitWebPage(u *url.URL) error {
 	req.Header.Set("Accept", "application/json")
 	var lm loginMethods
 	if err := v.client.Do(req, nil, &lm); err != nil {
+		if v.fallback != nil {
+			if err := v.fallback(u); err != nil {
+				return errgo.NoteMask(err, "form login not supported, fallback failed", errgo.Any)
```
rogpeppe: I think you should just return errgo.Mask(err, errgo.Any) here. We're setting up a fallback chain here, and I don't think that each link in the chain should prefix the error message from the next fallback.

httpbakery/form/form.go
mhilton: done

httpbakery/form/form.go
mhilton: done

httpbakery/form/form.go
```diff
@@ -123,9 +140,21 @@ func (v webPageVisitor) visitWebPage(u *url.URL) error {
 	req.Header.Set("Accept", "application/json")
 	var lm loginMethods
 	if err := v.client.Do(req, nil, &lm); err != nil {
```
rogpeppe, line 141: Thought for the future: if we have several kinds of supported non-interactive logins, it's somewhat wasteful to do a network round trip for each fallback when they'll all be wanting to fetch the same login methods page. We should think about ways we can avoid that.

httpbakery/form/form.go
mhilton: done

httpbakery/form/form.go
```diff
@@ -123,9 +140,21 @@ func (v webPageVisitor) visitWebPage(u *url.URL) error {
...
+				return errgo.NoteMask(err, "form login not supported, fallback failed", errgo.Any)
+			}
+			return nil
+		}
 		return errgo.Notef(err, "cannot get login methods")
 	}
 	if lm.Form == "" {
+		if v.fallback != nil {
+			if err := v.fallback(u); err != nil {
+				return errgo.NoteMask(err, "form login not supported, fallback failed", errgo.Any)
```
rogpeppe: ditto

mhilton: done

httpbakery/form/form.go
mhilton, line 141: Interesting, although so far I haven't seen a need to chain to anything other that open a browser. It is probable it will happen.

rogpeppe:
LGTM with some minor suggestions, thanks.

urosj:
:+1: did not do QA, but i think you had another branch to test this with a standalone test

PR #73, opened 2015-09-30
httpbakery: work around go issue #12796

rogpeppe:
We ensure that the tardy reads made by http.Request.Do cannot harm
our readers.

### DIFF
--- a/httpbakery/client.go
+++ b/httpbakery/client.go
@@ -10,6 +10,7 @@ import (
 	"net/http/cookiejar"
 	"net/url"
 	"strings"
+	"sync"
 
 	"github.com/juju/loggo"
 	"golang.org/x/net/publicsuffix"
@@ -343,14 +344,47 @@ func (c *Client) setRequestBody(req *http.Request, body io.ReadSeeker) error {
 	if body == nil {
 		return nil
 	}
-	if req.Body == nil {
-		req.Body = ioutil.NopCloser(body)
-	} else {
-		_, err := body.Seek(0, 0)
-		if err != nil {
+	if req.Body != nil {
+		req.Body.Close()
+		if _, err := body.Seek(0, 0); err != nil {
 			return errgo.Notef(err, "cannot seek to start of request body")
 		}
 	}
+	// Always replace the body with a new readStopper so that
+	// the old request cannot interfere with the new request's reader.
+	req.Body = &readStopper{
+		r: body,
+	}
+	return nil
+}
+
+var errClosed = errgo.New("reader has been closed")
+
+// readStopper works around an issue with the net/http
+// package (see http://golang.org/issue/12796).
+// Because the first HTTP request might not have finished
+// reading from its body when it returns, we need to
+// ensure that the second request does not race on Read,
+// so this type implements a Reader that prevents all Read
+// calls to the underlying Reader after Close has been called.
+type readStopper struct {
+	mu sync.Mutex
+	r  io.ReadSeeker
+}
+
+func (r *readStopper) Read(buf []byte) (int, error) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	if r.r == nil {
+		return 0, errClosed
+	}
+	return r.r.Read(buf)
+}
+
+func (r *readStopper) Close() error {
+	r.mu.Lock()
+	r.r = nil
+	r.mu.Unlock()
 	return nil
 }
 
--- a/httpbakery/client_test.go
+++ b/httpbakery/client_test.go
@@ -142,6 +142,52 @@ func (s *ClientSuite) TestRepeatedRequestWithBody(c *gc.C) {
 	c.Assert(bodyReader.byteCount, gc.Equals, len(bodyText)*2)
 }
 
+func (s ClientSuite) TestWithLargeBody(c *gc.C) {
+	// This test is designed to fail when run with the race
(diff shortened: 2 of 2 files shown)


httpbakery/client.go
```diff
@@ -343,14 +344,47 @@ func (c *Client) setRequestBody(req *http.Request, body io.ReadSeeker) error {
 	if body == nil {
 		return nil
 	}
-	if req.Body == nil {
-		req.Body = ioutil.NopCloser(body)
-	} else {
-		_, err := body.Seek(0, 0)
-		if err != nil {
+	if req.Body != nil {
+		req.Body.Close()
```
cmars, line 348: Even though that req.Body is probably going to be a `readStopper`, I'd check the error and at least log it, just in case it ever isn't.

httpbakery/client.go
```diff
@@ -343,14 +344,47 @@ func (c *Client) setRequestBody(req *http.Request, body io.ReadSeeker) error {
...
+		if _, err := body.Seek(0, 0); err != nil {
 			return errgo.Notef(err, "cannot seek to start of request body")
 		}
 	}
+	// Always replace the body with a new readStopper so that
+	// the old request cannot interfere with the new request's reader.
+	req.Body = &readStopper{
```
cmars, line 355: Concerned about this expectation set by the `net/http#NewRequest` godoc:

> If the provided body is also an io.Closer, the returned Request.Body is set to body and will be closed by the Client methods Do, Post, and PostForm, and Transport.RoundTrip.

If I use `httpbakery/Client.DoWithBody` with a body that's an `os.File`, how would that file now get closed? Would the caller now need to close the file? That might be surprising, if the caller is expecting net/http-like behavior.

I think using `Close()` to reset the reference to the underlying reader is a pretty clever workaround for the [net/http bug](https://github.com/golang/go/issues/12796)... but should we also keep track of the underlying reader across repeated calls to `DoWithBody` so that we can eventually close it for the caller?

[Edited]

httpbakery/client.go
rogpeppe, line 348: We know statically that it is a readStopper (an earlier version of the code actually did req.Body.(*readStopper).stop()) and therefore cannot return an error. Even if it wasn't, there's nothing useful that the error can tell us from a Reader - it's not like we're checking whether previously buffered writes can fail. If the Close fails, so what? FWIW the net/http package also ignores the error from Request.Body.Close.

httpbakery/client.go
rogpeppe, line 355: That paragraph from the NewRequest godoc is not relevant to us here, because we don't allow a Body to be passed in the request (we actually check and return an error if that's the case). It would be surprising if the body was closed, as the body passed to DoWithBody is not a Closer.

The NewRequest behaviour is an acknowledged bad design - it just cannot be changed due to backward-compatibility concerns.

httpbakery/client.go
```diff
@@ -343,14 +344,47 @@ func (c *Client) setRequestBody(req *http.Request, body io.ReadSeeker) error {
...
+// reading from its body when it returns, we need to
+// ensure that the second request does not race on Read,
+// so this type implements a Reader that prevents all Read
+// calls to the underlying Reader after Close has been called.
+type readStopper struct {
+	mu sync.Mutex
+	r  io.ReadSeeker
+}
+
+func (r *readStopper) Read(buf []byte) (int, error) {
```
alesstimec, line 375: missing godoc.

httpbakery/client.go
```diff
@@ -343,14 +344,47 @@ func (c *Client) setRequestBody(req *http.Request, body io.ReadSeeker) error {
...
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	if r.r == nil {
+		return 0, errClosed
+	}
+	return r.r.Read(buf)
+}
+
+func (r *readStopper) Close() error {
```
alesstimec, line 384: missing godoc.

httpbakery/client.go
```diff
@@ -343,14 +344,47 @@ func (c *Client) setRequestBody(req *http.Request, body io.ReadSeeker) error {
...
 	}
+	// Always replace the body with a new readStopper so that
+	// the old request cannot interfere with the new request's reader.
+	req.Body = &readStopper{
+		r: body,
+	}
+	return nil
+}
+
+var errClosed = errgo.New("reader has been closed")
```
alesstimec, line 361: this error is only used in one place.. maybe we could return errgo.New("reader has been closed") in L381

alesstimec:
a few nitpicks but LGTM

httpbakery/client.go
rogpeppe, line 375: Given that the type is not exported, the functionality is trivial and described in the doc for the type, I'm not sure a doc comment here is worth it (or on Close below).

httpbakery/client.go
rogpeppe, line 361: I'd prefer to use a variable. It saves an allocation in the discharge case and this form is quite conventional.

mhilton:
I think I'd like to see a test that the readStopper works in the way you think. Otherwise LGTM

rogpeppe:
> I think I'd like to see a test that the readStopper works in the way you think. Otherwise LGTM

As discussed online, I think that the functional test should be enough - readStopper is just an implementation detail towards that fix.

PR #74, opened 2015-10-01
add caveats to InteractiveDischarger macaroons

mhilton:
Some tests require caveats to be added to the macaroons produced.

### DIFF
--- a/bakerytest/bakerytest.go
+++ b/bakerytest/bakerytest.go
@@ -132,9 +132,14 @@ func (d *Discharger) PublicKeyForLocation(loc string) (*bakery.PublicKey, error)
 	return nil, bakery.ErrNotFound
 }
 
+type dischargeResult struct {
+	err  error
+	cavs []checkers.Caveat
+}
+
 type discharge struct {
 	cavId string
-	c     chan error
+	c     chan dischargeResult
 }
 
 // InteractiveDischarger is a Discharger that always requires interraction to
@@ -214,7 +219,7 @@ func (d *InteractiveDischarger) checker(req *http.Request, cavId, cav string) ([
 	d.mu.Lock()
 	id := fmt.Sprintf("%d", d.id)
 	d.id++
-	d.waiting[id] = discharge{cavId, make(chan error, 1)}
+	d.waiting[id] = discharge{cavId, make(chan dischargeResult, 1)}
 	d.mu.Unlock()
 	visitURL := fmt.Sprintf("%s/visit?waitid=%s", d.Discharger.server.URL, id)
 	waitURL := fmt.Sprintf("%s/wait?waitid=%s", d.Discharger.server.URL, id)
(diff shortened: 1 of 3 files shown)


urosj:
:+1:

bakerytest/bakerytest.go
```diff
@@ -276,7 +276,7 @@ func (d *InteractiveDischarger) wait(w http.ResponseWriter, r *http.Request) {
 // particular interaction is complete. It causes any waiting requests to
 // return. If err is not nil then it will be returned by the
 // corresponding /wait request.
-func (d *InteractiveDischarger) FinishInteraction(w http.ResponseWriter, r *http.Request, err error) {
+func (d *InteractiveDischarger) FinishInteraction(w http.ResponseWriter, r *http.Request, cavs []checkers.Caveat, err error) {
```
rogpeppe, line 286: It's a bit of a pity that this is a backwardly incompatible change, but I guess it's not too bad because it's in test code not production code.

If we cared about backward compatibility I guess we could keep the old
method and add FinishInteractionWithCaveats, but as you said online,
the API is very new so probably no-one is using it.

bakerytest/bakerytest.go
```diff
@@ -286,8 +286,10 @@ func (d *InteractiveDischarger) FinishInteraction(w http.ResponseWriter, r *http
 		httprequest.WriteJSON(w, code, body)
 		return
 	}
+	discharge.err = err
+	discharge.cavs = cavs
 	select {
-	case discharge.c <- err:
+	case discharge.c <- struct{}{}:
```
rogpeppe: I think I'd prefer for the error and caveats to be sent on the channel rather than set in the discharge object because there's no necessary reason that FinishInteraction might not be called twice concurrently, which would cause a race condition here.

bakerytest/bakerytest_test.go
```diff
@@ -53,6 +53,9 @@ func (s *suite) TestDischargerSimple(c *gc.C) {
 }
 
 var failChecker = bakery.FirstPartyCheckerFunc(func(s string) error {
+	if s == "test pass" {
```
rogpeppe: This seems a bit of a stretch of the failChecker contract...

bakerytest/bakerytest_test.go
```diff
@@ -142,7 +145,11 @@ func (s *suite) TestInteractiveDischarger(c *gc.C) {
 	var d *bakerytest.InteractiveDischarger
 	d = bakerytest.NewInteractiveDischarger(nil, http.HandlerFunc(
 		func(w http.ResponseWriter, r *http.Request) {
-			d.FinishInteraction(w, r, nil)
+			d.FinishInteraction(w, r, []checkers.Caveat{
+				checkers.Caveat{
+					Condition: "test pass",
```
rogpeppe, line 147: Rather than overloading failChecker, whose sole purpose is to fail when checking caveats, I'd suggest using a custom checker below that only succeeds when it sees the condition specified here.

rogpeppe, line 147: Even better, make the checker save the condition tested so that you know that it's actually tested the condition you're expecting.

rogpeppe:
LGTM modulo some suggestions above, thanks.

PR #77, opened 2015-10-05
httpbakery: add Client.HandleError method

rogpeppe:
We want this available so that we can use the httpbakery discharge and cookie
setting logic for requests that are not necessarily made with Do or DoWithBody
(for example websocket connections).


mhilton:
:+1: LGTM

alesstimec:
LGTM

PR #79, opened 2015-10-07
httpbakery: do not include method and URL in Do error

rogpeppe:
Higher level packages (e.g. httprequest) have a habit of adding this information,
so it's often redundant to have it added here too.


mhilton:
:+1: LGTM

mattyw:
LGTM

PR #82, opened 2015-10-22
httpbakery: add PublicKeyRing

rogpeppe:
Also move the existing PublicKeyForLocation function and tests into the same files.

### DIFF
--- a/httpbakery/keyring.go
+++ b/httpbakery/keyring.go
@@ -1,0 +1,105 @@
+package httpbakery
+
+import (
+	"encoding/json"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"strings"
+
+	"gopkg.in/errgo.v1"
+
+	"gopkg.in/macaroon-bakery.v1/bakery"
+)
+
+// NewPublicKeyRing returns a new public keyring that uses
+// the given client to find public keys and uses the
+// given cache as a backing. If cache is nil, a new
+// cache will be created. If client is nil, http.DefaultClient will
+// be used.
+func NewPublicKeyRing(client *http.Client, cache *bakery.PublicKeyRing) *PublicKeyRing {
+	if cache == nil {
+		cache = bakery.NewPublicKeyRing()
+	}
+	if client == nil {
+		client = http.DefaultClient
+	}
+	return &PublicKeyRing{
+		client: client,
+		cache:  cache,
+	}
+}
+
+// PublicKeyRing represents a public keyring that can interrogate
+// remote services for their public keys. By default it refuses
+// to use insecure URLs.
+type PublicKeyRing struct {
+	client        *http.Client
+	allowInsecure bool
+	cache         *bakery.PublicKeyRing
+}
+
+// AllowInsecure allows insecure URLs. This can be useful
+// for testing purposes.
+func (kr *PublicKeyRing) AllowInsecure() {
+	kr.allowInsecure = true
+}
+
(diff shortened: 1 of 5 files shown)


httpbakery/keyring.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	k, err := kr.cache.PublicKeyForLocation(loc)
+	if err == nil {
+		return k, nil
+	}
+	k, err = PublicKeyForLocation(kr.client, loc)
+	if err != nil {
+		return nil, errgo.Mask(err)
+	}
+	kr.cache.AddPublicKeyForLocation(loc, true, k)
```
alesstimec: err = kr.cache.AddPublicKeyForLocation(...)
if err != nil {
}

httpbakery/keyring_test.go
```diff
@@ -0,0 +1,131 @@
...
+	jujutesting.LoggingSuite
+}
+
+var _ = gc.Suite(&KeyringSuite{})
+
+func (s *KeyringSuite) TestCachePrepopulated(c *gc.C) {
+	cache := bakery.NewPublicKeyRing()
+	key, err := bakery.GenerateKey()
+	c.Assert(err, gc.IsNil)
+	cache.AddPublicKeyForLocation("https://0.1.2.3/", true, &key.Public)
```
alesstimec, line 29: check the returned error

alesstimec:
LGTM after the two comments above are addressed..

httpbakery/keyring.go
```diff
@@ -0,0 +1,97 @@
...
+func NewPublicKeyRing(client *http.Client, cache *bakery.PublicKeyRing) *PublicKeyRing {
+	if cache == nil {
+		cache = bakery.NewPublicKeyRing()
+	}
+	if client == nil {
+		client = http.DefaultClient
+	}
+	return &PublicKeyRing{
+		client: client,
+		cache:  cache,
```
alesstimec, line 29: i know bool defaults to false, but it might be nice to be explicit here
allowInsecure: false,

rogpeppe, line 29: I think leaving it as omitted is reasonable. The sense of the boolean is that way for a reason.

httpbakery/keyring.go
rogpeppe: done (even though it actually can never happen)

httpbakery/keyring_test.go
rogpeppe, line 29: if it fails, the test will fail, so I don't care much.

httpbakery/keyring.go
```diff
@@ -0,0 +1,101 @@
...
+		// loc is an invalid URL which we have already checked.
+		return nil, errgo.Notef(err, "cannot cache discharger URL %q", loc)
+	}
+	return k, nil
+}
+
+// PublicKeyForLocation returns the public key from a macaroon
+// discharge server running at the given location URL.
+// Note that this is insecure if an http: URL scheme is used.
+func PublicKeyForLocation(client *http.Client, url string) (*bakery.PublicKey, error) {
```
mhilton, line 80: It might be nice if you could use nil for the client parameter and have it use http.DefaultClient

mhilton:
:+1: LGTM with a small thought.

httpbakery/keyring.go
rogpeppe, line 80: done

PR #83, opened 2015-11-02
bakery: make non-prefix matching on PublicKeyRing slash-suffix-insensitive

rogpeppe:
This means we can avoid prefix matching (which we actually don't want in most circumstances) but still make the public key ring work regardless of whether the configuration file specifies a trailing slash or not.


cmars:
LGTM

rogpeppe:
This isn't actually necessary in practice, as pointed out mhilton.

PR #85, opened 2015-12-01
httpbakery/client: If we cannot discharge but the cause is InteractionError then pass this back as the cause

mattyw:


alesstimec:
LGTM

rogpeppe:
LGTM except the PR description is wrong. The error is InteractionError (an error
when interacting) not InteractionRequired, which is quite a different thing.

PR #86, opened 2015-12-10
Return cached public key for http location.

cmars:
Allow public key lookup to succeed for previously cached HTTP locations.
The 'insecure' setting should only prevent requests for the public key
from unencrypted endpoints.


alesstimec:
LGTM!

PR #87, opened 2016-01-19
httpbakery: allow servers to set desired cookie name

rogpeppe:
This will allow us to mitigate cookie buildup in clients.

### DIFF
--- a/httpbakery/client.go
+++ b/httpbakery/client.go
@@ -313,6 +313,9 @@ func (c *Client) HandleError(reqURL *url.URL, err error) error {
 		return errgo.Notef(err, "cannot make cookie")
 	}
 	cookie.Path = cookiePath
+	if name := respErr.Info.CookieNameSuffix; name != "" {
+		cookie.Name = "macaroon-" + name
+	}
 	c.Jar.SetCookies(reqURL, []*http.Cookie{cookie})
 	return nil
 }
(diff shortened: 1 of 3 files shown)


httpbakery/client.go
```diff
@@ -313,6 +313,9 @@ func (c *Client) HandleError(reqURL *url.URL, err error) error {
 		return errgo.Notef(err, "cannot make cookie")
 	}
 	cookie.Path = cookiePath
+	if name := respErr.Info.CookieName; name != "" {
```
mhilton: I would argue that this should be MacaroonName, because apart from anything else it's not the actual cookie name.

httpbakery/error.go
```diff
@@ -77,6 +77,13 @@ type ErrorInfo struct {
 	// the original URL from which the error was returned.
 	MacaroonPath string `json:",omitempty"`
 
+	// CookieName holds the desired cookie name
+	// suffix to be associated with the macaroon.
+	// The actual name used will be ("macaroon-" + CookieName).
+	// Clients may ignore this field - older clients will
+	// always use ("macaroon-" + macaroon.Signature() in hex).
+	CookieName string	`json:",omitempty"`
```
mhilton: MacaroonName ? (see above)

mhilton:
:+1: with a thought.

httpbakery/client.go
rogpeppe: I wondered about this earlier and thought that MacaroonName might be confusing because it's not really the name of the macaroon at all (and macaroons have an id which is really their name).

CookieNameSuffix?
MacaroonCookieName?
MacaroonCookieNameSuffix?

httpbakery/client.go
```diff
@@ -313,6 +313,9 @@ func (c *Client) HandleError(reqURL *url.URL, err error) error {
...
+		cookie.Name = "macaroon-" + name
```
urosj, line 317: Why do we have to prefix it with macaroon?

httpbakery/client.go
alesstimec: CookieNameSuffix  +1

alesstimec:
LGTM

httpbakery/client.go
rogpeppe, line 317: Because the code for checking macaroon cookies checks only the cookies with a "macaroon-" prefix, and we want this code to work against current APIs. If we didn't prefix the name here, then servers that didn't provide a name with the prefix wouldn't work.

httpbakery/client.go

PR #89, opened 2016-02-26
httpbakery: add CheckRequestM

rogpeppe:
We want to know the macaroons that were used to check the request,
and this lets us do that. Also add bakery.Service.CheckAnyM for the
same reason.


jrwren:
:+1:

mhilton:
:+1:

PR #90, opened 2016-03-09
Use external webbrowser.Open helper.

frankban:
Also make the package compile without errors.

### DIFF
--- a/httpbakery/browser.go
+++ b/httpbakery/browser.go
@@ -4,58 +4,23 @@ import (
 	"fmt"
 	"net/url"
 	"os"
-	"os/exec"
-	"runtime"
-	"strings"
-)
 
-var browser = map[string]string{
-	"linux":   "sensible-browser",
-	"darwin":  "open",
-	"freebsd": "xdg-open",
-	"netbsd":  "xdg-open",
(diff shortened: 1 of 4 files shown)


httpbakery/browser.go
```diff
@@ -4,58 +4,22 @@ import (
...
-		cmd := exec.Command(args[0], args[1:]...)
-		cmd.Stdout = os.Stdout
-		cmd.Stderr = os.Stderr
-		cmd.Start()
-		go cmd.Wait()
-	} else {
+		return nil
+	}
+	if err == webbrowser.ErrNoBrowser {
 		fmt.Fprintf(os.Stderr, "Please open this URL in your browser to authorize:\n%s\n", url)
```
rogpeppe, line 22: return nil?
I think in this case it's not useful to return an error.

rogpeppe:
LGTM, thanks.
Please rebase before merging.

httpbakery/browser.go
frankban, line 22: Yeah, definitely forgot to add the "return nil" here.

mhilton:
:+1:

PR #91, opened 2016-03-10
bakery: fix broken tests

rogpeppe:
By using a fixed id in the test, we were associating several root keys with the
same id. Fix by using a different id in each test.


frankban:
:+1: the tests pass now. Thank you!

bakery/service.go
```diff
@@ -103,6 +103,7 @@ func (svc *Service) PublicKey() *PublicKey {
 // describes the error (other errors might be returned in other
 // circumstances).
 func (svc *Service) Check(ms macaroon.Slice, checker FirstPartyChecker) error {
+	logger.Infof("check with store %T %#v", svc.store)
```
mhilton, line 106: This seems like a debugging message to me is Infof too high a logging level?

PR #93, opened 2016-03-14
httpbakery/form: expose LoginWithForm

mhilton:
This allows more complex VisitWebPage functions to be built which
make decisions about which login method to use after retrieving the
login methods.


frankban:
:+1:

mhilton:
This is very out of date, and almost certainly not wanted now.

PR #94, opened 2016-03-15
httpbakery: add support for interaction method requests

rogpeppe:
This formalises the interaction method request
protocol that already exists into httpbakery, making
it (hopefully) straightforward to mix and match
different kinds of discharge fulfillment interactions.

We also update the form package (backwardly incompatible
but we hope no-one's looking) to use the new
scheme. The agent package will be similarly updated
in a subsequent PR.

### DIFF
--- a/httpbakery/browser.go
+++ b/httpbakery/browser.go
@@ -24,3 +24,17 @@ func OpenWebBrowser(url *url.URL) error {
 	}
 	return err
 }
+
+// WebBrowserVisitor holds an interactor that supports the "Interactive"
+// method by opening a web browser at the required location.
+var WebBrowserVisitor Visitor = webBrowserVisitor{}
+
+type webBrowserVisitor struct{}
+
+func (webBrowserVisitor) VisitWebPage(client *Client, methodURLs map[string]*url.URL) error {
+	u := methodURLs[UserInteractionMethod]
+	if u == nil {
+		return ErrMethodNotSupported
+	}
+	return OpenWebBrowser(u)
+}
--- a/httpbakery/form/form.go
+++ b/httpbakery/form/form.go
@@ -2,10 +2,10 @@
 package form
 
 import (
-	"net/http"
 	"net/url"
 
 	"github.com/juju/httprequest"
+	"github.com/juju/loggo"
 	"golang.org/x/net/publicsuffix"
 	"gopkg.in/errgo.v1"
 	"gopkg.in/juju/environschema.v1"
@@ -13,6 +13,8 @@ import (
 
 	"gopkg.in/macaroon-bakery.v1/httpbakery"
 )
+
+var logger = loggo.GetLogger("httpbakery.form")
 
 /*
 PROTOCOL
@@ -51,49 +53,12 @@ client's responsibility to interpret the schema and present it to the
 user.
 */
 
-// SetUpAuth configures form authentication on c. The VisitWebPage field
-// in c will be set to a function that will attempt form-based
-// authentication using f to perform the interaction with the user and
-// fall back to using the current value of VisitWebPage if form-based
-// authentication is not supported.
-func SetUpAuth(c *httpbakery.Client, f form.Filler) {
-	c.VisitWebPage = VisitWebPage(c, f, c.VisitWebPage)
-}
-
-// VisitWebPage creates a function suitable for use with
-// httpbakery.Client.VisitWebPage. The new function downloads the schema
-// from the specified server and calls f.Fill. The map returned by f.Fill
-// should match the schema specified, but this is not verified before
-// sending to the server. Any errors returned by f.Fill or fallback will
-// not have their cause masked.
-//
-// If the new function detects that form login is not supported by the
-// server and fallback is not nil then fallback will be called to perform
-// the visit.
-func VisitWebPage(d httprequest.Doer, f form.Filler, fallback func(u *url.URL) error) func(u *url.URL) error {
-	v := webPageVisitor{
-		client: &httprequest.Client{
-			Doer: d,
-		},
-		filler:   f,
-		fallback: fallback,
-	}
-	return v.visitWebPage
-}
-
-// webPageVisitor contains the state required by visitWebPage.
-type webPageVisitor struct {
-	client   *httprequest.Client
-	filler   form.Filler
-	fallback func(u *url.URL) error
-}
-
-// loginMethods contains the response expected from the login URL. It
-// only checks for the "form" method as that is the only one that can be
-// handled.
-type loginMethods struct {
-	Form string `json:"form"`
-}
+const (
+	// InteractionMethod is the methodURLs key
+	// used for a URL that can be used for form-based
+	// interaction.
+	InteractionMethod = "form"
+)
 
 // SchemaRequest is a request for a form schema.
 type SchemaRequest struct {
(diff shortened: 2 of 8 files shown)


httpbakery/form/form.go
```diff
@@ -51,49 +53,12 @@ client's responsibility to interpret the schema and present it to the
...
-// only checks for the "form" method as that is the only one that can be
-// handled.
-type loginMethods struct {
-	Form string `json:"form"`
-}
+const (
+	// InteractionMethod is the methodURLs key
+	// used for a URL that can be used for form-based
+	// interaction.
+	InteractionMethod = "form"
```
cmars, line 60: Would we ever use something other than a form post for HTTP interaction?

cmars:
I can see where this is going -- an abstraction over the client discharge acquisition pattern -- but I wonder if this is getting too abstract. What interactions do you have in mind specifically, other than HTTP form POSTs? Maybe I lack context on the motivation for this patch.

I'm concerned with backward compatibility as 16.04 approaches. It sounds like that's been considered -- but is it tested? I'd feel better about this if it was targeting a v2-unstable branch so that v1 clients could be imported and compatibility tested.

httpbakery/form/form.go
mhilton, line 60: form in this case doesn't mean a form post, but a download of a form schema from the discharger to use to interact with the user. The only known client of this is what it is being modified for, so I don't think backwards compatibilty is a major concern there.

httpbakery/browser.go
```diff
@@ -24,3 +25,21 @@ func OpenWebBrowser(url *url.URL) error {
...
 	return err
 }
+
+// WebBrowserVisitor holds an interactor that supports the "Interactive"
+// method by opening a web browser at the required location.
+var WebBrowserVisitor = webBrowserVisitor{}
+
+type webBrowserVisitor struct{}
+
+func (webBrowserVisitor) VisitWebPage(client *Client, methodURLs map[string]string) error {
```
mhilton: This has the wrong signature now so it isn't actually a Visitor.

httpbakery/form/form.go
```diff
@@ -117,48 +82,48 @@ type LoginBody struct {
...
-	if err := v.client.Do(req, nil, &lm); err != nil {
-		if v.fallback != nil {
-			if err := v.fallback(u); err != nil {
-				return errgo.Mask(err, errgo.Any)
-			}
-			return nil
-		}
-		return errgo.Notef(err, "cannot get login methods")
+func (v Visitor) VisitWebPage(client *httpbakery.Client, methodURLs map[string]*url.URL) error {
+	return v.visitWebPage(client, methodURLs)
```
mhilton, line 99: client.Client?

httpbakery/visitor.go
```diff
@@ -0,0 +1,108 @@
...
+	VisitWebPage(client *Client, methodURLs map[string]*url.URL) error
+}
+
+const (
+	// UserInteractionMethod is the methodURLs key
+	// used for a URL that should be visited in a user's
+	// web browser. This is also the URL that can
+	// be used to fetch the available login methods (with
+	// an appropriate Accept header).
+	UserInteractionMethod = "Interactive"
```
mhilton: To be compatible with IdM this should be spelled "interactive" (although currently I think IdM is spelling it "usso").

```
$ bhttp  https://api.jujucharms.com/identity/v1/login "Accept:application/json"
{
    agent: "https://api.jujucharms.com/identity/v1/idp/agent/agent"
    usso: "https://login.ubuntu.com/+openid?openid.claimed_id=http%3A%2F%2Fspecs.openid.net%2Fauth%2F2.0%2Fidentifier_select\u0026openid.identity=http%3A%2F%2Fspecs.openid.net%2Fauth%2F2.0%2Fidentifier_select\u0026openid.mode=checkid_setup\u0026openid.ns=http%3A%2F%2Fspecs.openid.net%2Fauth%2F2.0\u0026openid.realm=https%3A%2F%2Fapi.jujucharms.com%2Fidentity%2Fv1%2Fidp%2Fusso%2Fcallback\u0026openid.return_to=https%3A%2F%2Fapi.jujucharms.com%2Fidentity%2Fv1%2Fidp%2Fusso%2Fcallback\u0026openid.lp.query_membership=blues-development%2Ccharm-beta\u0026openid.ns.lp=http%3A%2F%2Fns.launchpad.net%2F2007%2Fopenid-teams\u0026openid.ns.sreg=http%3A%2F%2Fopenid.net%2Fextensions%2Fsreg%2F1.1\u0026openid.sreg.required=email%2Cfullname%2Cnickname"
    usso_oauth: "https://api.jujucharms.com/identity/v1/idp/usso_oauth/oauth"
}
```

Anyway it should be lowercase to match everything else (and how IdM used to work).

httpbakery/visitor.go
```diff
@@ -0,0 +1,108 @@
...
+
+func (v multiVisitor) VisitWebPage(client *Client, methodURLs map[string]*url.URL) error {
+	u := methodURLs[UserInteractionMethod]
+	if u == nil {
+		return errgo.Newf("cannot get interaction methods because no %q URL found", UserInteractionMethod)
+	}
+	if urls, err := GetInteractionMethods(client, u); err == nil {
+		// We succeeded in getting the set of interaction methods from
+		// the discharger. Use them.
+		methodURLs = urls
```
mhilton, line 67: I would like to check here that methodURLs contains UserInteractionMathods and merge it in if it doesn't. This mkaes the whole thing backwards compatible with the currently deployed IdM.

mhilton:
:+1: after a few fixes.

httpbakery/visitor.go
```diff
@@ -0,0 +1,108 @@
...
+// specifying that the associated URL should be opened
+// in a web browser for the user to interact with.
+//
+// See FallbackVisitor for a way to gain access to
+// alternative methods.
+//
+// A Visitor implementation may return ErrMethodNotSupported
+// if it cannot handle any of the supplied methods.
+type Visitor interface {
+	VisitWebPage(client *Client, methodURLs map[string]*url.URL) error
```
alesstimec, line 28: may i propose a new name for this method: DoInteraction.. i think this has outgrown it's original purpose - to use a web browser to open a web page..

httpbakery/visitor.go
```diff
@@ -0,0 +1,108 @@
...
+
+import (
+	"net/http"
+	"net/url"
+
+	"github.com/juju/httprequest"
+	"gopkg.in/errgo.v1"
+)
+
+var ErrMethodNotSupported = errgo.New("interaction method not supported")
```
alesstimec, line 13: missing godoc

httpbakery/visitor.go
```diff
@@ -0,0 +1,108 @@
...
+// discharger for available methods and then tries each
+// of the given visitors in turn until one succeeds
+// or fails with an error cause other than ErrMethodNotSupported.
+func NewMultiVisitor(methods ...Visitor) Visitor {
+	return &multiVisitor{
+		supportedMethods: methods,
+	}
+}
+
+func (v multiVisitor) VisitWebPage(client *Client, methodURLs map[string]*url.URL) error {
```
alesstimec, line 56: godoc, please

httpbakery/visitor.go
rogpeppe: Oh, I'll fix that then, and make it always allow the interactive method.

httpbakery/visitor.go
rogpeppe, line 67: done

httpbakery/visitor.go
rogpeppe, line 28: I see where you're coming from, and I did initially start by
doing something similar. But I moved on to thinking that some name with "Visit" in
is nice, because it goes together with the associated VisitURL in the
error. We could just call the method Visit, I guess, but that doesn't
seem quite descriptive enough somehow. What do you think?

alesstimec:
LGTM... though i'm a bit worried, that the it's getting harder to maintain backwards compatibility.. might be time to move to V2 soon

httpbakery/form/form.go
rogpeppe, line 99: I think it's reasonable that we use the bakery client to do the interaction.
Can you think of a reason why we wouldn't want to?

httpbakery/browser.go
rogpeppe: fixed

PR #96, opened 2016-04-01
httpbakery: Pass through errors in MulitVisitor

mhilton:
If a WebPageVisitor that is part of a MultiVisitor returns an error
the cause is not masked by MultiVisitor.

### DIFF
--- a/httpbakery/visitor.go
+++ b/httpbakery/visitor.go
@@ -52,7 +52,9 @@ func NewMultiVisitor(methods ...Visitor) Visitor {
 
 // VisitWebPage implements Visitor.VisitWebPage by obtaining all the
 // available interaction methods and calling v.supportedMethods until it
-// finds one that recognizes the method.
+// finds one that recognizes the method. If a Visitor returns an error
+// other than ErrMethodNotSupported the error will be immediately
+// returned to the caller; its cause will not be masked.
 func (v multiVisitor) VisitWebPage(client *Client, methodURLs map[string]*url.URL) error {
 	// The Client implementation will always include a UserInteractionMethod
 	// entry taken from the VisitURL field in the error, so use that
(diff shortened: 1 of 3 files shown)


httpbakery/visitor.go
```diff
@@ -52,7 +52,9 @@ func NewMultiVisitor(methods ...Visitor) Visitor {
 
 // VisitWebPage implements Visitor.VisitWebPage by obtaining all the
 // available interaction methods and calling v.supportedMethods until it
-// finds one that recognizes the method.
+// finds one that recognizes the method. If a Visitor returns an error
+// other than ErrMethodNotSupported the error will be immediately
+// returned to the caller, its cause will not be masked.
```
rogpeppe: s/,/;/

httpbakery/visitor_test.go
```diff
@@ -147,6 +148,29 @@ func (*VisitorSuite) TestUserInteractionFallback(c *gc.C) {
...
+		fmt.Fprint(w, `{"method": "http://somewhere/something"}`)
+	}))
+	defer srv.Close()
+
+	testError := errgo.New("test error")
+	called := 0
+	v := httpbakery.NewMultiVisitor(
+		visitorFunc(func(_ *httpbakery.Client, m map[string]*url.URL) error {
+			c.Check(m["method"], gc.NotNil)
+			called++
```
rogpeppe: No real need to check called here - if we get a testError out of VisitWebPage, there's nowhere else it could've come from. Similarly for checking m["method"] - we're not testing that aspect of MultiVisitor here.

rogpeppe:
LGTM with a couple of trivial suggestions.

httpbakery/visitor.go
mhilton: done

httpbakery/visitor_test.go
mhilton: done

urosj:
LGTM

PR #98, opened 2016-05-03
httpbakery: ensure no read after DoWithBody

rogpeppe:
This juju bug https://bugs.launchpad.net/juju-core/+bug/1576509
shows that reading after a request has returned, which
http.Client.Do can (see Go issue https://github.com/golang/go/issues/12796) is a problem.

Ensure that DoWithBody can never do that by closing
the existing readStopper body which is in place for
just that reason.


urosj:
Addresses issue: https://github.com/golang/go/issues/15521

urosj:
LGTM

mhilton:
:+1:

PR #99, opened 2016-05-23
bakery: add RootKeyStorage

rogpeppe:
This makes it easy to re-use macaroon root keys as desired,
and also to use domain-object-specific root keys to
mint macaroons that are limited in scope.

Even though we re-use public keys, we ensure that all macaroon
ids are unique by appending a nonce to the id.

A subsequent PR will introduce a mongodb-backed implementation of RootKeyStorage.


bakery/service.go
```diff
@@ -10,6 +10,7 @@ import (
 	"strings"
 
 	"github.com/juju/loggo"
+	"github.com/rogpeppe/fastuuid"
```
fabricematrat, line 13: fast because there is a slow one :)

bakery/service.go
```diff
@@ -18,11 +19,14 @@ import (
 
 var logger = loggo.GetLogger("bakery")
 
+var uuidGen = fastuuid.MustNewGenerator()
```
fabricematrat, line 22: Does it mean it will panic here ?

bakery/storage.go
```diff
@@ -24,8 +28,71 @@ type Storage interface {
...
+func NewMemRootKeyStorage() RootKeyStorage {
+	return new(memRootKeyStorage)
+}
+
+type memRootKeyStorage struct {
+	mu  sync.Mutex
+	key []byte
+}
+
+func (s *memRootKeyStorage) Get(id string) ([]byte, error) {
```
fabricematrat, line 69: doc ?

bakery/storage.go
```diff
@@ -24,8 +28,71 @@ type Storage interface {
...
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if id != "0" || s.key == nil {
+		return nil, ErrNotFound
+	}
+	return s.key, nil
+}
+
+func (s *memRootKeyStorage) RootKey() (rootKey []byte, id string, err error) {
```
fabricematrat, line 80: doc ?

bakery/service.go
```diff
@@ -213,17 +251,48 @@ func (svc *Service) NewMacaroon(id string, rootKey []byte, caveats []checkers.Ca
...
-	// TODO look at the caveats for expiry time and associate
-	// that with the storage item so that the storage can
-	// garbage collect it at an appropriate time.
-	if err := svc.store.Put(m.Id(), &storageItem{
-		RootKey: rootKey,
-	}); err != nil {
-		return nil, fmt.Errorf("cannot save macaroon to store: %v", err)
+	if svc.rkStore == nil {
+		if err := svc.store.Put(m.Id(), rootKey); err != nil {
+			return nil, fmt.Errorf("cannot save macaroon to store: %v", err)
```
fabricematrat: I wonder why we use fmt.Errorf as we have a logger declared.

fabricematrat:
:+1:

bakery/service.go
mhilton, line 22: Only if it can't read from /dev/random. at which point we can't do much anyway.

bakery/service.go
mhilton: fmt.Errorf creates a new error.

bakery/storage.go
```diff
@@ -24,8 +28,71 @@ type Storage interface {
...
+	// Note that the root keys should remain available for as long
+	// as the macaroons using them are valid.
+	//
+	// Note that there is no need for it to return a new root key
+	// for every call - keys may be reused, although some key
+	// cycling is over time is advisable.
+	RootKey() (rootKey []byte, id string, err error)
+}
+
+// NewFallbackStorage returns a RootKeyStorage implementation
```
mhilton: This seems to be part of another PR

mhilton:
:+1: with a few thoughts

bakery/service.go
rogpeppe, line 13: fast 'cos it's a lot faster than just reading from /dev/rand :)

bakery/storage.go
rogpeppe, line 69: done

bakery/storage.go
rogpeppe, line 80: done

bakery/service.go
rogpeppe: I'll change it to use errgo.Notef.

bakery/service.go
mattyw, line 22: Even so, macaroon-bakery is just a library, it should be up to the app to decide, but at least it would happen on startup I guess. One concern I have is that depending on the un knowable logging strategy of said app this panic could be logged without any context of where it came from. Why not just do it in NewService?

bakery/service.go
```diff
@@ -40,6 +44,12 @@ type NewServiceParams struct {
 	// an in-memory storage will be used.
 	Store Storage
 
+	// RootKeyStore is used to store macaroon information. If this
```
mattyw: s/macaroon information/root keys ?

bakery/service.go
```diff
@@ -79,8 +93,27 @@ func NewService(p NewServiceParams) (*Service, error) {
 	return svc, nil
 }
 
+// WithRootKeyStore returns a copy of service where macaroon creation and
+// lookup uses the given store to look up macaroon root keys
```
mattyw: given root key store. Service has a Store attribute so I'd like to be explicit here

bakery/service.go
```diff
@@ -79,8 +93,27 @@ func NewService(p NewServiceParams) (*Service, error) {
...
+// get root keys and ids.
+//
+// When NewMacaroon is called on the returned Service,
+// is must always be called with an empty id and rootKey.
```
mattyw: s/is/it

bakery/service.go
```diff
@@ -79,8 +93,27 @@ func NewService(p NewServiceParams) (*Service, error) {
...
+func (svc *Service) WithRootKeyStore(store RootKeyStorage) *Service {
+	svc1 := *svc
+	svc1.rkStore = store
```
mattyw, line 104: maybe also

```
svc1.Store = nil
```

mattyw:
:+1: at least one typo, other comment nitpicks and assorted other biksheds

bakery/service.go
rogpeppe: done

bakery/service.go
rogpeppe: done ish

bakery/service.go
rogpeppe: done

bakery/service.go
rogpeppe, line 104: done

PR #100, opened 2016-05-26
mgostorage: add RootKeyStorage support

rogpeppe:
This means that we can use the same underlying collection for
both long-lived and short-lived macaroon root keys, and use
the cache for almost all accesses.

### DIFF
--- a/bakery/mgostorage/rootkey.go
+++ b/bakery/mgostorage/rootkey.go
@@ -1,0 +1,348 @@
+package mgostorage
+
+import (
+	"crypto/rand"
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/juju/loggo"
+	"gopkg.in/errgo.v1"
+	"gopkg.in/mgo.v2"
+	"gopkg.in/mgo.v2/bson"
+
+	"gopkg.in/macaroon-bakery.v1/bakery"
+)
+
+// Functions defined as variables so they can be overidden
+// for testing.
+var (
+	timeNow = time.Now
+
+	mgoCollectionFindId = (*mgo.Collection).FindId
+)
+
+var logger = loggo.GetLogger("bakery.mgostorage")
+
+// maxPolicyCache holds the maximum number of storage policies that can
+// hold cached keys in a given RootKeys instance.
+//
+// 100 is probably overkill, given that practical systems will
+// likely only have a small number of active policies on any given
+// macaroon collection.
+const maxPolicyCache = 100
+
+// RootKeys represents a cache of macaroon root keys.
+type RootKeys struct {
+	maxCacheSize int
+
+	// TODO (rogpeppe) use RWMutex instead of Mutex here so that
+	// it's faster in the probably-common case that we
+	// have many contended readers.
+	mu       sync.Mutex
+	oldCache map[string]rootKey
+	cache    map[string]rootKey
+
+	// current holds the current root key for each storage policy.
+	current map[Policy]rootKey
+}
+
+type rootKey struct {
+	Id      string `bson:"_id"`
+	Created time.Time
+	Expires time.Time
+	RootKey []byte
+}
+
+// isValid reports whether the root key contains a key. Note that we
+// always generate non-empty root keys, so we use this to find
+// whether the root key is empty or not.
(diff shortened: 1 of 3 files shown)


bakery/mgostorage/rootkey.go
```diff
@@ -0,0 +1,344 @@
...
+
+	mgoCollectionFindId = (*mgo.Collection).FindId
+)
+
+var logger = loggo.GetLogger("bakery.mgostorage")
+
+// maxPolicyCache holds the maximum number
+// of storage policies that can hold cached keys in
+// a given RootKeys instance.
+// In practice, this could probably be 3 or 4 and it
```
mhilton: I'm not sure this line tells me anything. Should I change this number to 3 or 4?

rogpeppe: Adjusted comment.

mhilton:
:+1:

bakery/mgostorage/rootkey.go
```diff
@@ -0,0 +1,348 @@
...
+
+	mgoCollectionFindId = (*mgo.Collection).FindId
+)
+
+var logger = loggo.GetLogger("bakery.mgostorage")
+
+// maxPolicyCache holds the maximum number of storage policies that can
+// hold cached keys in a given RootKeys instance.
+//
+// This value is probably overkill, given that practical systems will
```
mattyw: 100 is probably overkill

bakery/mgostorage/rootkey.go
```diff
@@ -0,0 +1,348 @@
...
+// likely only have a small number of active policies on any given
+// macaroon collection.
+const maxPolicyCache = 100
+
+// RootKeys represents a cache of macaroon root keys.
+type RootKeys struct {
+	maxCacheSize int
+
+	// TODO use RWMutex instead of Mutex here so that
```
mattyw: TODO (rogpeppe)

bakery/mgostorage/rootkey.go
```diff
@@ -0,0 +1,348 @@
...
+	oldCache map[string]rootKey
+	cache    map[string]rootKey
+
+	// current holds the current root key for each storage policy.
+	current map[Policy]rootKey
+}
+
+type rootKey struct {
+	Id      string `bson:"_id"`
+	Created time.Time
```
mattyw, line 52: I'm fond of the convention that we tag all attributes, YMMV

rogpeppe, line 52: MM does. :)

bakery/mgostorage/rootkey.go
```diff
@@ -0,0 +1,348 @@
...
+		// space.
+		s.current = make(map[Policy]rootKey)
+	}
+	s.current[policy] = key
+}
+
+type rootKeyStorage struct {
+	keys   *RootKeys
+	policy Policy
+	coll   *mgo.Collection
```
mattyw, line 214: I feel like we shouldn't be holding on to a collection like this.  We can take a collection name as string and accept a session that we use for connecting. Otherwise we'll need to create a new rootKeyStorage when our connection to mongo drops

rogpeppe, line 214: We can create a new RootKeyStorage instance for every
request - that's very cheap to do and each one will use the same underlying
cache (the RootKeys instance).

cmars:
> Why use the same collection for both? Why not just slap a TTL index on the authentication macaroon collection, and not have one on the longer-lived authorization macaroons?

We considered that, but that a) constrains you to exactly two policies (long term and short term) and also it means that when looking up macaroons, you'd need to look up in two collections to be sure a macaroon doesn't exist. And logically, all the macaroons here are part of the same authorization domain - both long term and short term macaroons can be used to authorize the same set of operations.

FWIW in the charm store, we have three policies (long term, very short term and medium term).

bakery/mgostorage/rootkey.go
rogpeppe: done

bakery/mgostorage/rootkey.go
rogpeppe: done

mattyw:
:+1:

PR #101, opened 2016-06-08
httpbakery: use latest version rather than zero

rogpeppe:
This means that when we move to using a later version,
old servers will use the latest version they support rather
than falling back to version 0.

We should make sure that all servers incorporate this change
before starting to migrate to the new bakery/macaroons standards.


urosj:
LGTM. Trying to match with the latest that is supported by a running service sounds ok, even if it fails. Certainly better then going to 0.

mhilton:
:+1:

rogpeppe:
:shipit:

PR #102, opened 2016-06-21
bakery: add third-party caveat id encoding 0

mhilton:
Also rearrange some code to make adding encodings easier.

### DIFF
--- a/bakery/codec.go
+++ b/bakery/codec.go
@@ -4,11 +4,19 @@ import (
 	"bytes"
 	"crypto/rand"
 	"encoding/base64"
+	"encoding/binary"
 	"encoding/json"
-	"fmt"
 
 	"golang.org/x/crypto/nacl/box"
+
+	"gopkg.in/errgo.v1"
 )
+
+type caveatInfo struct {
+	peerPublicKey *PublicKey
+	rootKey       []byte
+	condition     string
+}
 
 type caveatIdRecord struct {
 	RootKey   []byte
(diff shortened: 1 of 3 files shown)


bakery/codec.go
```diff
@@ -5,11 +5,16 @@ import (
...
-	"fmt"
 
-	"golang.org/x/crypto/nacl/box"
+	"gopkg.in/errgo.v1"
 )
 
+type caveatInfo struct {
+	PeerPublicKey *PublicKey
+	RootKey       []byte
+	Predicate     string
```
mattyw: why not just Condition just to make it horifically obvious?

bakery/codec.go
```diff
@@ -23,109 +28,158 @@ type caveatId struct {
...
+	v0NonceOffset                     = v0FirstPartyPublicKeyOffset + KeyLen
+	v0EncryptedPartOffset             = v0NonceOffset + NonceLen
+
+	v0RootKeyOffset    = 1
+	v0CaveatTextOffset = v0RootKeyOffset + 24
+)
+
+// encodeCaveatIdV0 creates a version 0 third-party caveat.
+func encodeCaveatIdV0(key *KeyPair, ci caveatInfo) ([]byte, error) {
+	var nonce [NonceLen]byte
```
rogpeppe, line 77: nonce, err := randomBytes(NonceLen)

bakery/codec.go
```diff
@@ -23,109 +28,158 @@ type caveatId struct {
...
-	if d.key == nil {
-		return nil, fmt.Errorf("no public key for caveat id decryption")
+// decodeCaveatIdV0 decodes a version 0 caveat id.
+func decodeCaveatIdV0(key *KeyPair, id []byte) (caveatInfo, error) {
+	if len(id) < v0EncryptedPartOffset {
+		return caveatInfo{}, errgo.New("caveat id too short")
 	}
-	if !bytes.Equal(d.key.Public.Key[:], id.ThirdPartyPublicKey.Key[:]) {
-		return nil, fmt.Errorf("public key mismatch")
+	if !bytes.Equal(key.Public.Key[:4], id[v0ThirdPartyPublicKeyPrefixOffset:v0FirstPartyPublicKeyOffset]) {
```
mattyw: this 4 is the same 4 as in 

```
v0ThirdPartyPublicKeyPrefixOffset + 4
```

is that right? Should that be a const as well or is that too much const?

bakery/codec.go
```diff
@@ -5,11 +5,16 @@ import (
...
 	"encoding/base64"
 	"encoding/json"
-	"fmt"
 
-	"golang.org/x/crypto/nacl/box"
+	"gopkg.in/errgo.v1"
 )
 
+type caveatInfo struct {
+	PeerPublicKey *PublicKey
```
rogpeppe: Do these fields warrant export?

bakery/codec_test.go
```diff
@@ -0,0 +1,61 @@
...
+
+	gc "gopkg.in/check.v1"
+)
+
+type codecSuite struct{}
+
+var _ = gc.Suite(&codecSuite{})
+
+func (*codecSuite) TestJSONRoundTrip(c *gc.C) {
+	fpkp, err := GenerateKey()
```
mattyw: If you were looking for other suggestions for this bikeshed, you could  make a MustGenerateKey that panics on err, that would remove some code in the various tests that generate keys

bakery/codec.go
```diff
@@ -23,109 +28,158 @@ type caveatId struct {
...
+	v0CaveatTextOffset = v0RootKeyOffset + 24
+)
+
+// encodeCaveatIdV0 creates a version 0 third-party caveat.
+func encodeCaveatIdV0(key *KeyPair, ci caveatInfo) ([]byte, error) {
+	var nonce [NonceLen]byte
+	if _, err := rand.Read(nonce[:]); err != nil {
+		return nil, errgo.Notef(err, "cannot generate random number for nonce")
+	}
+	sealed := encodeSecretPartV0(key, ci, &nonce)
```
rogpeppe: I don't think I'd bother with all the offset calculations.

I'd suggest something like:

```
data := make([]byte, 0, 1+4+KeyLen+NonceLen+len(sealed))
data = append(data, 0)      // version
data = append(data, ci.PeerPublicKey.Key[0:4]...)
data = append(data, key.Public.Key[:]...)
data = append(data, nonce...)
data = append(data, sealed...)
```

bakery/codec.go
```diff
@@ -23,109 +28,158 @@ type caveatId struct {
...
+	v0ThirdPartyPublicKeyPrefixOffset = 1
+	v0FirstPartyPublicKeyOffset       = v0ThirdPartyPublicKeyPrefixOffset + 4
+	v0NonceOffset                     = v0FirstPartyPublicKeyOffset + KeyLen
+	v0EncryptedPartOffset             = v0NonceOffset + NonceLen
+
+	v0RootKeyOffset    = 1
+	v0CaveatTextOffset = v0RootKeyOffset + 24
+)
+
+// encodeCaveatIdV0 creates a version 0 third-party caveat.
```
rogpeppe, line 68: I'd add a comment describing the format:

// Format:
// 
// version 0 [1 byte]
// first 4 bytes of third-party Curve25519 public key [4 bytes]
// first-party Curve25519 public key [32 bytes]
// nonce [24 bytes]
// encrypted secret part [rest of message]

bakery/codec.go
```diff
@@ -23,109 +28,158 @@ type caveatId struct {
...
+	return data, nil
 }
 
-// newBoxDecoder creates a new BoxDecoder using the given key pair.
-func newBoxDecoder(key *KeyPair) *boxDecoder {
-	return &boxDecoder{
-		key: key,
+// encodeSEcretPartV0 creates an encrypted secret part of the third party
+// caveat at version 0.
+func encodeSecretPartV0(key *KeyPair, ci caveatInfo, nonce *[NonceLen]byte) []byte {
```
rogpeppe: Again, a comment describing the format would be nice:

// Format of encrypted blob:
// 
// version 0 [1 byte]
// root key [24 bytes]
// UTF-8 encoded caveat id text [rest of message]

bakery/codec.go
```diff
@@ -23,109 +28,158 @@ type caveatId struct {
...
+	data := make([]byte, len(ci.RootKey)+len(ci.Predicate)+1)
```
rogpeppe: Less obvious a win here, but probably nice to use the same append idiom I suggested above:

```
data := make([]byte, 0, 1 + len(ci.RootKey) + len(ci.Predicate))
data = append(data, 0)    // version
data = append(data, ci.RootKey...)
data = key.seal(data, ci.Predicate...)
return key.seal(ci.PeerPublicKey, nonce, data)
```

rogpeppe: You could even do the whole thing with two allocs rather
than three if you leveraged the fact that box.Seal can append to an existing slice.

bakery/codec.go
```diff
@@ -23,109 +28,158 @@ type caveatId struct {
...
+		Predicate: record.Condition,
+	}, nil
 }
 
-func (d *boxDecoder) encryptedCaveatId(id caveatId) ([]byte, error) {
-	if d.key == nil {
-		return nil, fmt.Errorf("no public key for caveat id decryption")
+// decodeCaveatIdV0 decodes a version 0 caveat id.
+func decodeCaveatIdV0(key *KeyPair, id []byte) (caveatInfo, error) {
+	if len(id) < v0EncryptedPartOffset {
```
rogpeppe: Again, I think the offsets don't make things much clearer here.

I think I'd suggest something like:

```
    if len(id) < 1+ 4+KeyLen+NonceLen+box.Overhead {
        return errgo.New("caveat id too short")
    }

    myKeyPrefix := id[0:4]
    if !bytes.Equal(key.Public[0:4], myKeyPrefix) {
        ...
    }
    id = id[4:]

    var peerPublicKey PublicKey
    copy(peerPublicKey[:], id)
    id = id[KeyLen:]

    var nonce [NonceLen]byte
    copy(nonce[:], id)
    id = id[NonceLen:]

    data, err := key.open(&peerPublicKey, &nonce, id)
```

bakery/codec.go
```diff
@@ -23,109 +28,158 @@ type caveatId struct {
...
 	if err != nil {
-		return nil, fmt.Errorf("cannot base64-decode encrypted caveat id: %v", err)
+		return caveatInfo{}, errgo.Notef(err, "cannot decrypt caveat id %q", id)
 	}
-	out, ok := box.Open(nil, sealed, &nonce, id.FirstPartyPublicKey.boxKey(), d.key.Private.boxKey())
-	if !ok {
-		return nil, fmt.Errorf("decryption of public-key encrypted caveat id %#v failed", id)
+	if data[0] != 0 {
+		return caveatInfo{}, errgo.Newf("unknown caveat version %d", data[0])
 	}
```
rogpeppe, line 207: We need to check the decrypted length here, otherwise we might get a panic.

bakery/keys.go
```diff
@@ -125,6 +125,20 @@ func (key *KeyPair) String() string {
 	return key.Public.String()
 }
 
+// seal encrypts data using peerPublicKey and signs it using key.Private.
+func (key *KeyPair) seal(peerPublicKey *PublicKey, nonce *[NonceLen]byte, data []byte) []byte {
+	return box.Seal(nil, data, nonce, peerPublicKey.boxKey(), key.Private.boxKey())
+}
+
+// open decrypts data using key.Private. and authenticates using peerPublicKey.
```
rogpeppe: s/. and/c/ and/

bakery/codec.go
```diff
@@ -23,109 +28,158 @@ type caveatId struct {
...
-	return record.RootKey, record.Condition, nil
+	cid, err := key.open(tpid.FirstPartyPublicKey, &nonce, secret)
+	if err != nil {
+		return caveatInfo{}, errgo.Notef(err, "cannot decrypt caveat id %#v", tpid)
+	}
+	var record caveatIdRecord
+	if err := json.Unmarshal(cid, &record); err != nil {
+		return caveatInfo{}, errgo.Notef(err, "cannot decode third party caveat record")
+	}
+	// Do not include the Peer Public Key as it was not returned
```
rogpeppe: I don't understand this comment. Which caller are we talking about here?

httpbakery/discharge.go
```diff
@@ -100,6 +101,18 @@ func (d *dischargeHandler) serveDischarge1(p httprequest.Params) (interface{}, e
...
+		// caveat ids of this type is unencoded so decode the
+		// base64 now.
+		id1, err := base64.URLEncoding.DecodeString(id)
+		if err != nil {
+			return nil, badRequestErrorf("cannot decode id: %s", err)
+		}
+		id = string(id1)
+	}
 	var resp dischargeResponse
 	m, err := d.svc.Discharge(bakery.ThirdPartyCheckerFunc(checker), id)
```
rogpeppe, line 104: Perhaps we should make Discharge take a []byte now?

bakery/codec_test.go
```diff
@@ -0,0 +1,61 @@
...
+	c.Assert(err, gc.IsNil)
+
+	res, err := decodeCaveatId(tpkp, cid)
+	c.Assert(err, gc.IsNil)
+	c.Assert(res.PeerPublicKey, gc.IsNil)
+	c.Assert(res.RootKey, gc.DeepEquals, rootKey[:])
+	c.Assert(res.Predicate, gc.Equals, "is-authenticated-user")
+}
+
+func (*codecSuite) TestV0RoundTrip(c *gc.C) {
```
rogpeppe: I'd like to have more tests for some of the error paths here - caveat ids too short, bad decryption, encrypted data too show, different version numbers, etc.

rogpeppe:
This looks great, thanks. A bunch of comments and suggestions, but nothing major.

httpbakery/discharge.go
mhilton, line 104: That would be a backwards incompatible API change (which might be necessary anyway, but I'd like to avoid if possible)

bakery/codec.go
mhilton: done (it's condition)

bakery/codec.go
mhilton, line 77: I'm leaving this as I would have to copy it out into an array to use it.

bakery/codec.go
mhilton: done

bakery/codec.go
mhilton: no, changed

bakery/codec_test.go
mhilton: I've moved these into a SetUpTest so the code only exists once.

bakery/codec.go
mhilton: done

bakery/codec.go
mhilton, line 68: done

bakery/codec.go
mhilton: done

bakery/codec.go
mhilton: done

bakery/codec.go
mhilton: done

bakery/codec.go
mhilton, line 207: done

bakery/keys.go
mhilton: These functions have been removed, so done

bakery/codec.go
mhilton: I've removed the comment, done

bakery/codec_test.go
mhilton: done

bakery/codec.go
```diff
@@ -23,109 +31,198 @@ type caveatId struct {
...
 }
 
-// boxDecoder decodes caveat ids for third-party service that were encoded to
-// the third-party with authenticated public key encryption compatible with
-// NaCl box.
-type boxDecoder struct {
-	key *KeyPair
+const (
+	publicKeyPrefixLen = 4
+	rootKeyLen         = 24
```
rogpeppe: This should go.

rogpeppe:
LGTM thanks!

PR #103, opened 2016-06-23
all: new v2-unstable version

rogpeppe:
We use the macaroon.v2-unstable package, allowing us to
have binary caveat ids and macaroon ids.

We change some of the API accordingly, removing some elements
that are unnecessary, such as the first party location argument
to the Discharge method.


rogpeppe:
Targetted to wrong branch.

PR #104, opened 2016-06-23
all: new v2-unstable version

rogpeppe:
We use the macaroon.v2-unstable package, allowing us to
have binary caveat ids and macaroon ids.

We change some of the API accordingly, removing some elements
that are unnecessary, such as the first party location argument
to the AcquireDischarge method.

Note that this changes the format of root keys stored by the mgostorage
package. There is some code to partially deal with that by falling back
to fetching from the old format, but a subsequent PR will make that more
comprehensive.

This also adds the ability for a third party discharger to be able to
see the keys used to decrypt the third party caveat, which will
enable some proposed use cases.

### DIFF
--- a/bakery/codec.go
+++ b/bakery/codec.go
@@ -12,12 +12,6 @@ import (
 	"gopkg.in/errgo.v1"
 )
 
-type caveatInfo struct {
-	peerPublicKey *PublicKey
-	rootKey       []byte
-	condition     string
-}
-
 type caveatIdRecord struct {
 	RootKey   []byte
 	Condition string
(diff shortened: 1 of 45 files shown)


mhilton:
:+1:

bakery/codec.go
```diff
@@ -32,22 +26,27 @@ type caveatId struct {
 }
 
 // encodeJSONCaveatId creates a JSON encoded third-party caveat.
-func encodeJSONCaveatId(key *KeyPair, ci caveatInfo) ([]byte, error) {
+func encodeJSONCaveatId(
```
urosj, line 33: Knowing the protocol, this makes sense, however, It would be nice to have an explanation why thirdPartyPub is needed.

bakery/codec.go
```diff
@@ -32,22 +26,27 @@ type caveatId struct {
...
+	condition string,
+	rootKey []byte,
+	thirdPartyPub *PublicKey,
```
urosj: s/thirdPartyPub/thirdPartyPubKey?

bakery/codec.go
```diff
@@ -61,9 +60,7 @@ func encodeJSONCaveatId(key *KeyPair, ci caveatInfo) ([]byte, error) {
 	return buf, nil
 }
 
-const (
-	publicKeyPrefixLen = 4
-)
+const publicKeyPrefixLen = 4
```
urosj: Why 4?

bakery/mgostorage/rootkey.go
```diff
@@ -226,7 +226,23 @@ func (s *rootKeyStorage) Get(id string) ([]byte, error) {
...
+func (s *rootKeyStorage) getFromMongo(id []byte) (rootKey, error) {
+	var key rootKey
+	err := mgoCollectionFindId(s.coll, id).One(&key)
+	if err != nil {
+		if err == mgo.ErrNotFound {
+			return s.getLegacyFromMongo(string(id))
+		}
+		return rootKey{}, errgo.Notef(err, "cannot get key from database")
+	}
+	// TODO migrate the key from the old format to the new format.
```
urosj, line 238: But old ones are handled as well and they should expire, why do we need a migration?

httpbakery/client.go
```diff
@@ -137,9 +137,8 @@ type Client struct {
 // discharge-acquisition process used by a Client.
 type DischargeAcquirer interface {
 	// AcquireDischarge should return a discharge macaroon for the given third
-	// party caveat. The firstPartyLocation holds the location of the original
-	// macaroon.
-	AcquireDischarge(firstPartyLocation string, cav macaroon.Caveat) (*macaroon.Macaroon, error)
+	// party caveat.
+	AcquireDischarge(cav macaroon.Caveat) (*macaroon.Macaroon, error)
```
urosj, line 141: An explanation why we can drop the location bit in the comment above, maybe?

urosj:
Some minor comments there. As mentioned, I find all the casting irritating. 

👍

httpbakery/client.go
rogpeppe, line 141: I think that belongs as a code review comment, not as a comment in the code, as there's no particular reason (apart from my misguided reasoning) why we should include the first party location anywhere at all.

bakery/mgostorage/rootkey.go
rogpeppe, line 238: Yeah, it _might_ be OK - I'm sure I thought of a reason earlier why we should do it.
I'll leave the TODO in for the moment.

bakery/codec.go
rogpeppe, line 33: done

bakery/codec.go
rogpeppe: done

PR #107, opened 2016-07-22
bakery, httpbakery: support v2 macaroons and caveat ids.

rogpeppe:
This has a bunch of changes which together
mean that the bakery supports binary-encoded
macaroons and caveat ids. This causes cookie
sizes to drop to around half their previous size (~1600 bytes
to ~800 bytes). That size will continue to drop in subsequent PRs.

Note that there is still one failing test which will require
an update to httprequest for a fix. Rather than delaying
this PR further, we intend to land this as an interim branch
and fix the discharge fallback behaviour in the next PR.

Rename PublicKeyLocator to ThirdPartyLocator and make
it convey information on the version of the third party too.

Remove PublicKeyRing as it's complex and has marginal use.

Use version 2 macaroons when possible, while staying backwardly
compatible with old clients and dischargers.

Remove functions added only because we couldn't change
the signature of existing functions (e.g. CheckAnyM)

Add version arguments to methods where the version is
important (e.g. bakery.Service.NewMacaroon).

Add version information to local third party caveat
so we know what kind of caveat encoding to use.

Changed bakerytest.NewDischarger to be more generally
useful, and added a helper to make it straightforward to
get parsed caveats as before.

Use httprequest to implement the HTTP discharge handler,
and implement the new endpoints needed to implement
the ThirdPartyLocator interface.


httpbakery/client_test.go
```diff
@@ -34,6 +33,8 @@ type ClientSuite struct {
 
 var _ = gc.Suite(&ClientSuite{})
 
+var macaroonCurrentVersion = bakery.MacaroonVersion(bakery.LatestVersion)
```
mhilton, line 36: Is this safe to do always? If it is then please add a comment explaining why, otherwise perhaps use a bakery.MacaroonVersion constant.

rogpeppe, line 36: Why would it not be safe to do always?

mhilton, line 36: Are the Macaroon version and bakery version always going to be in lockstep? in which case why have two versions?

mhilton:
:+1: with a small thought

httpbakery/client_test.go
rogpeppe, line 36: The MacaroonVersion function derives the correct macaroon version given a the bakery version. They're not in lockstep, which is why we call the function to do this (and why macaroonCurrentVersion is a variable not a constant). Were you thinking this was just a type conversion, perhaps?

mhilton, line 36: Yes It looked like a type conversion, please ignore.

httpbakery/error.go
```diff
@@ -275,24 +269,30 @@ func newDischargeRequiredErrorWithVersion(m *macaroon.Macaroon, path string, ori
...
-// request. If the protocol cannot be determined, or is invalid,
-// the original version of the protocol is used.
-func versionFromRequest(req *http.Request) version {
+// RequestVersion determines the bakery protocol version from a client
+// request. If the protocol cannot be determined, or is invalid, the
+// original version of the protocol is used. If a later version is
+// found, the latest known version is used, which is OK because versions
+// are backwardly compatible.
+//
+// TODO as there are no known version 0 clients, default to version 1
```
jrwren, line 278: Why not go ahead and do this?

jrwren:
👍  But I must admit this is a lot of code and I feel overwhelmed reviewing it. More eyes might be better.

bakery/codec.go
```diff
@@ -85,25 +107,25 @@ func encodeCaveatIdV0(
 		return nil, errgo.Notef(err, "cannot generate random number for nonce")
 	}
 	data := make([]byte, 0, 1+publicKeyPrefixLen+KeyLen+NonceLen+1+binary.MaxVarintLen64+len(rootKey)+len(condition)+box.Overhead)
-	data = append(data, 0) //version
+	data = append(data, byte(Version2)) //version
```
mattyw, line 110: you seem to make a big struct and then append to it. If you're doing an initial large make can't you put all these in the right index?

bakery/keys.go
```diff
@@ -79,25 +78,53 @@ func (k *Key) UnmarshalText(text []byte) error {
...
+	// Version holds latest the bakery protocol version supported
+	// by the discharger.
+	Version Version
+}
+
+// ThirdPartyLocator is used to find information on third
+// party discharge services.
+type ThirdPartyLocator interface {
+	// ThirdPartyInfo returns information on the third
+	// party at the given location. It returns ErrNotFound if no match is found.
```
mattyw, line 96: Comments like this trouble me slightly. You can't say for certain that it returns ErrNotFound here. But you can say that it should do.

bakery/keys.go
```diff
@@ -79,25 +78,53 @@ func (k *Key) UnmarshalText(text []byte) error {
...
+	ThirdPartyInfo(loc string) (ThirdPartyInfo, error)
+}
+
+// ThirdPartyLocatorMap implements a simple ThirdPartyLocator.
+// A trailing slash on locations is ignored.
+type ThirdPartyLocatorStore struct {
+	m map[string]ThirdPartyInfo
+}
+
+func NewThirdPartyLocatorStore() *ThirdPartyLocatorStore {
```
mattyw, line 108: // NewThirdPartyLocatorStore returns a new instance of ThirdPartyLocatorStore

bakery/service.go
```diff
@@ -136,25 +167,53 @@ func (svc *Service) Check(ms macaroon.Slice, checker FirstPartyChecker) error {
...
+		// so we assume that it's the base64 encoding of a new-style
+		// macaroon id, so we base64 decode it.
+		//
+		// Note that old-style ids always start with an ASCII character >= 4
+		// (> 32 in fact) so this logic won't be triggered for those.
+		dec := make([]byte, base64.RawURLEncoding.DecodedLen(len(id)))
+		n, err := base64.RawURLEncoding.Decode(dec, id)
+		if err == nil {
+			// Set the id only on success - if it's a bad encoding, we'll get a not-found error
+			// which is fine.
```
mattyw: which is fine because...?

httpbakery/keyring.go
```diff
@@ -1,105 +1,113 @@
...
+	switch resp.StatusCode {
+	case http.StatusOK:
+		var r dischargeInfoResponse
+		if err := httprequest.UnmarshalJSONResponse(resp, &r); err != nil {
+			return bakery.ThirdPartyInfo{}, errgo.Mask(err)
+		}
+		info.PublicKey = *r.PublicKey
+		info.Version = r.Version
+	case http.StatusNotFound:
+		// TODO(rog) this fallback does not work because
```
mattyw, line 99: I trust you have relevant cards on your board to fix this and the above when the latest httprequest lands

mattyw:
:+1: some comments from me, mostly trivial

bakery/codec.go
rogpeppe, line 110: It's cleaner and safer doing it this way. You don't need to add all the offsets cumulatively and if you happen to have got the initial size wrong, the only penalty is an allocation, not a panic.

bakery/keys.go
rogpeppe, line 96: I think it's reasonable to document an interface like this, for the user rather than the implementor of the interface. io.Reader (for example, as a canonical example of an interface) documents itself in this way and I think it's better than using "should" and "must" all over the place.

httpbakery/error.go
rogpeppe, line 278: Because that would affect a bunch of the tests and this PR is way too big already...

bakery/keys.go
rogpeppe, line 108: done (and a couple of other related doc comments too)

bakery/service.go
rogpeppe: done

httpbakery/keyring.go
rogpeppe, line 99: of course... :)

PR #110, opened 2016-11-11
bakery/checkers: refactor to use contexts and namespaces

rogpeppe:
We change the way that first party caveat checking works
so that instead of requiring a closure or object to be created
containing the request context (thus requiring a checker
to be created afresh for each request), we use an context.Context
to hold information used for verification by checkers.

This means we can treat checkers as much more static objects.
The idea is that any application will probably have only one
instance of a checker object with all the checks supported
by first party caveats in that application.

We also implement namespaces (as discussed in the Hague sprint).
A checkers.Caveat now contains a Namespace field containing a URI
of the namespace to use. This is resolved by a Namespace instance
when actually adding the caveat to the macaroon.

The delete the idservice example because it's a pain to fix and
we want to rewrite it to use the upcoming auth model anyway.

Note: this is a change to the v2-unstable branch - we are definitely
not preserving backward compatibility here!


bakery/checkers/checkers.go
```diff
@@ -4,192 +4,185 @@ package checkers
...
-	CondDeny         = "deny"
 )
 
-// ErrCaveatNotRecognized is the cause of errors returned
-// from caveat checkers when the caveat was not
-// recognized.
-var ErrCaveatNotRecognized = errgo.New("caveat not recognized")
+// Func is the type of a function used by Checker to check a caveat.
+// The cond will hold the condition that was registered and
+// the argument any additional caveat argument text.
```
mhilton: This is a slightly clunky sentence, can you see if you can reword it please?

bakery/checkers/checkers.go
```diff
@@ -4,192 +4,185 @@ package checkers
...
-	// Check checks that the given caveat holds true.
-	// The condition and arg are as returned
-	// from ParseCaveat.
-	//
-	// For a checker with an empty condition, a
-	// return of bakery.ErrCaveatNotRecognised from
-	// this method indicates that the condition was
-	// not recognized.
-	Check(cond, arg string) error
+// NewEmpty returns a checker  using the given namespace
```
mhilton: remove the double space.

bakery/checkers/checkers.go
```diff
@@ -4,192 +4,185 @@ package checkers
...
-// as well as bakery.FirstPartyChecker, calls to New can be nested.
-// For example, a checker can be easily added to an existing
-// MultiChecker, by doing:
-//
-//	checker := checkers.New(old, another)
-func New(checkers ...Checker) *MultiChecker {
-	return &MultiChecker{
-		checkers: checkers,
+// RegisterStd registers all the standard checkers in the given checker.
+// The standard namespace is also added to ns if not present.
```
mhilton: what's ns? you should probably say it is added without a prefix.

bakery/checkers/checkers.go
```diff
@@ -4,192 +4,185 @@ package checkers
...
-	// Always check for the error caveat so that we're
-	// sure to get a nice error message even when there
-	// are no other checkers. This also prevents someone
-	// from inadvertently overriding the error condition.
-	if cond == CondError {
-		return errBadCaveat
+// Register registers the given condition in the given namespace URI
+// to be checked with the given check function.
+// It will panic if the namespace is not registered or
+// if the condition has been registered twice.
```
mhilton: ... if the condition is already registered?

bakery/checkers/checkers_test.go
```diff
@@ -2,33 +2,36 @@ package checkers_test
...
+
+// testClock is a Clock implementation that always returns the above time.
+type testClock struct{}
+
+func (testClock) Now() time.Time {
 	return now
-}()
+}
+
+// FirstPartyCaveatChecker is declared here so we can avoid a cyclic
```
mhilton, line 30: I'm interested to know why this is necessary as we're in the test package.

bakery/checkers/declared.go
```diff
@@ -76,33 +77,42 @@ func (c Declared) Check(_, arg string) error {
...
-				conflicts = append(conflicts, key)
-				continue
-			}
-			info[key] = val
 		}
 	}
+	return InferDeclaredFromConditions(ns, conditions)
+}
+
+// InferDeclaredFromCaveats is like InferDeclared except that
```
mhilton: InferDeclaredFromConditions

bakery/checkers/namespace.go
```diff
@@ -0,0 +1,94 @@
...
+	// TODO compatibility
+	return ns.Resolve(uri)
+}
+
+// Resolve resolves the given schema URI to its registered prefix and
+// returns the prefix and whether the resolution was successful.
+//
+// If ns is nil, it is treated as if it were empty.
+//
+// ResolveCaveat does not mutate ns and may be called concurrently
```
mhilton: Resolve?

bakery/checkers/operations.go
```diff
@@ -0,0 +1,99 @@
...
+	"fmt"
+	"strings"
+
+	"golang.org/x/net/context"
+	errgo "gopkg.in/errgo.v1"
+)
+
+type opKey struct{}
+
+// ContextWithDeclared returns a context which is associated with all the
```
mhilton: ContextWithOperations

bakery/example/targetservice.go
```diff
@@ -50,36 +52,23 @@ func targetService(endpoint, authEndpoint string, authPK *bakery.PublicKey) (htt
...
 		srv.writeError(w, req, "gold", err)
 		return
 	}
 	fmt.Fprintf(w, "all is golden")
 }
 
 func (srv *targetServiceHandler) serveSilver(w http.ResponseWriter, req *http.Request) {
-	checker := srv.checkers(req, "silver")
-	if _, err := httpbakery.CheckRequest(srv.svc, req, nil, checker); err != nil {
+	ctxt := checkers.ContextWithOperations(context.TODO(), "gold")
```
mhilton: "silver"?

bakery/service.go
```diff
@@ -618,17 +642,27 @@ func (c ThirdPartyCheckerFunc) CheckThirdPartyCaveat(info *ThirdPartyCaveatInfo)
 	return c(info)
 }
 
-// FirstPartyChecker holds a function that checks first party caveats
-// for validity.
+// FirstPartyCaveatChecker is used to check first party caveats
+// for validity with respect to information in the provided context.
 //
 // If the caveat kind was not recognised, the checker should return
```
mhilton, line 648: This should be checkers.ErrCaveatNotRecognised and probably isn;t worth documenting here.

httpbakery/checkers.go
```diff
@@ -4,38 +4,66 @@ import (
...
 	"gopkg.in/errgo.v1"
 
 	"gopkg.in/macaroon-bakery.v2-unstable/bakery/checkers"
 )
 
-type httpContext struct {
-	req *http.Request
+type httpRequestKey struct{}
+
+func ContextWithRequest(ctxt context.Context, req *http.Request) context.Context {
```
mhilton, line 18: I'm not certain adding the whole request to the context is a good idea. How about something like?

``` go
func ContextFromRequest(parentt context.Context, req *http.Request) context.Context {
    ctx := parent
    if ctx == nil {
        ctx = req.Context() // or context.Background() depending on go version
    }
    ctx = context.WithValue(ctx, ipAddrKey, req.RemoteAddr)
    ctx = context.WithValue(ctx, originKey, req.Header.Get("Origin"))
    return ctx
}
```

mhilton (approved):
This is fantastic stuff, thanks

bakery/checkers/checkers.go
```diff
@@ -4,192 +4,185 @@ package checkers
...
 	"fmt"
-	"net"
+	"sort"
 	"strings"
 
+	"golang.org/x/net/context"
 	"gopkg.in/errgo.v1"
 )
 
+const StdNamespace = "std"
```
alesstimec, line 15: godoc, please

bakery/checkers/checkers.go
```diff
@@ -4,192 +4,185 @@ package checkers
...
-	//
-	// For a checker with an empty condition, a
-	// return of bakery.ErrCaveatNotRecognised from
-	// this method indicates that the condition was
-	// not recognized.
-	Check(cond, arg string) error
+// NewEmpty returns a checker  using the given namespace
+// that has no registered checkers.
+// If ns is nil, a new one will be created.
+func NewEmpty(ns *Namespace) *Checker {
```
alesstimec, line 62: NewEmptyChecker, perhaps?

bakery/checkers/checkers.go
```diff
@@ -4,192 +4,185 @@ package checkers
...
-		if err := c.Check(cond, arg); err != nil {
-			if checkerCond == "" && errgo.Cause(err) == ErrCaveatNotRecognized {
-				continue
-			}
-			return errgo.Mask(err, errgo.Any)
-		}
-		checked = true
+	prefix, ok := c.namespace.Resolve(uri)
+	if !ok {
+		panic(fmt.Errorf("no prefix registered for namespace %q when registering condition %q", uri, cond))
```
alesstimec, line 108: don't panic.. return an error?

bakery/checkers/checkers.go
```diff
@@ -245,101 +220,20 @@ func ErrorCaveatf(f string, a ...interface{}) Caveat {
...
+type checkerInfoByName []CheckerInfo
 
-// operationCaveat is a helper for AllowCaveat and DenyCaveat. It checks
-// that all operation names are valid before createing the caveat.
-func operationCaveat(cond string, op []string) Caveat {
-	for _, o := range op {
-		if strings.IndexByte(o, ' ') != -1 {
-			return ErrorCaveatf("invalid operation name %q", o)
-		}
+func (c checkerInfoByName) Less(i, j int) bool {
```
alesstimec, line 236: godoc

bakery/checkers/checkers.go
```diff
@@ -245,101 +222,20 @@ func ErrorCaveatf(f string, a ...interface{}) Caveat {
...
-		case CondAllow:
-			f := strings.Fields(arg)
-			if len(f) == 0 {
-				return errgo.New("no operations allowed")
-			}
-			return errgo.Newf("%s not allowed", f[0])
-		default:
-			return ErrCaveatNotRecognized
-		}
+func (c checkerInfoByName) Less(i, j int) bool {
```
alesstimec, line 236: godoc

bakery/checkers/checkers.go
```diff
@@ -245,101 +222,20 @@ func ErrorCaveatf(f string, a ...interface{}) Caveat {
...
 }
 
-// OperationChecker checks any allow or deny caveats, ensuring they do not
-// prohibit the named operation.
-type OperationChecker string
-
-// Condition implements Checker.Condition.
-func (OperationChecker) Condition() string {
-	return ""
+func (c checkerInfoByName) Swap(i, j int) {
```
alesstimec, line 244: godoc

bakery/checkers/checkers.go
```diff
@@ -245,101 +222,20 @@ func ErrorCaveatf(f string, a ...interface{}) Caveat {
...
-			}
-		}
-		if found == expect {
-			return nil
-		}
-		return fmt.Errorf("%s not allowed", o)
-	default:
-		return ErrCaveatNotRecognized
-	}
+func (c checkerInfoByName) Len() int {
```
alesstimec, line 248: godoc

bakery/service.go
```diff
@@ -618,17 +642,27 @@ func (c ThirdPartyCheckerFunc) CheckThirdPartyCaveat(info *ThirdPartyCaveatInfo)
...
+	// is valid with respect to the given context information.
+	CheckFirstPartyCaveat(ctxt context.Context, caveat string) error
+
+	// Namespace returns the namespace associated with the
+	// caveat checker.
+	Namespace() *checkers.Namespace
 }
 
-type FirstPartyCheckerFunc func(caveat string) error
+type FirstPartyCaveatCheckerFunc func(ctxt context.Context, caveat string) error
```
alesstimec: godoc

bakery/service.go
```diff
@@ -618,17 +642,27 @@ func (c ThirdPartyCheckerFunc) CheckThirdPartyCaveat(info *ThirdPartyCaveatInfo)
...
 
-func (c FirstPartyCheckerFunc) CheckFirstPartyCaveat(caveat string) error {
-	return c(caveat)
+func (c FirstPartyCaveatCheckerFunc) CheckFirstPartyCaveat(ctxt context.Context, caveat string) error {
```
alesstimec: godoc

bakery/service.go
```diff
@@ -618,17 +642,27 @@ func (c ThirdPartyCheckerFunc) CheckThirdPartyCaveat(info *ThirdPartyCaveatInfo)
...
+	return c(ctxt, caveat)
+}
+
+func (c FirstPartyCaveatCheckerFunc) Namespace() *checkers.Namespace {
```
alesstimec: godoc

httpbakery/checkers.go
```diff
@@ -4,38 +4,81 @@ import (
...
-// It does not include the "declared" checker, as that
-// must be added for each individual set of macaroons
-// that are checked.
-func Checkers(req *http.Request) checkers.Checker {
-	c := httpContext{req}
-	return checkers.Map{
-		checkers.CondClientIPAddr: c.clientIPAddr,
-		checkers.CondClientOrigin: c.clientOrigin,
+const (
+	CondClientIPAddr = "client-ip-addr"
```
alesstimec, line 30: godocs

httpbakery/checkers.go
```diff
@@ -4,38 +4,81 @@ import (
...
+	CondClientOrigin = "origin"
+)
+
+const CheckersNamespace = "http"
```
alesstimec, line 38: godoc

httpbakery/discharge.go
```diff
@@ -186,6 +189,10 @@ type dischargeResponse struct {
 	Macaroon *macaroon.Macaroon `json:",omitempty"`
 }
 
+// TODO remove the need for this by acquiring the namespace
```
mattyw, line 192: TODO (rog)

bakery/checkers/checkers.go
```diff
@@ -4,192 +4,187 @@ package checkers
...
-	// from ParseCaveat.
-	//
-	// For a checker with an empty condition, a
-	// return of bakery.ErrCaveatNotRecognised from
-	// this method indicates that the condition was
-	// not recognized.
-	Check(cond, arg string) error
-}
+const (
+	CondNeedDeclared = "need-declared"
```
mattyw, line 29: Why doesn't this get to live with the other consts?

bakery/checkers/checkers.go
```diff
@@ -4,192 +4,187 @@ package checkers
...
-// Condition implements Checker.Condition.
-func (f CheckerFunc) Condition() string {
-	return f.Condition_
-}
+var errBadCaveat = errgo.New("bad caveat")
 
-// Check implements Checker.Check
-func (f CheckerFunc) Check(cond, arg string) error {
-	return f.Check_(cond, arg)
+func checkError(ctxt context.Context, _, arg string) error {
```
mattyw, line 163: I'd be tempted (to the point of giving in to it) to move this function and the above var errBadCaveat closer to var allCheckers

bakery/checkers/checkers.go
```diff
@@ -4,192 +4,187 @@ package checkers
...
-func (m Map) Check(cond, arg string) error {
-	f, ok := m[cond]
-	if !ok {
-		return ErrCaveatNotRecognized
-	}
-	if err := f(cond, arg); err != nil {
-		return errgo.Mask(err, errgo.Any)
+func firstParty(cond, arg string) Caveat {
+	return Caveat{
+		Condition: cond + " " + arg,
```
mattyw: I'd be tempted to make this a function
func condFormat(c, a string) string { return c + " " + a }

mattyw: Alternatively - you also create a Caveat in httpCaveat with a different namespace, which is currently the only other place cond + " " + args appears. Maybe replace with a NewCaveat(cond, arg string, ns *Namespace)

mattyw: I know it seems nitpicky, but I like standard things being in as few places as possible

bakery/checkers/checkers_test.go
mattyw: Also interested - what's the reason?

bakery/checkers/checkers_test.go
```diff
@@ -219,47 +140,46 @@ var checkerTests = []struct {
...
-				return errWrongArg
-			}
-			return nil
-		},
+// argChecker returns a checker function that checks
+// that the caveat condition is checkArg.
+func argChecker(expectCond, checkArg string) checkers.Func {
+	return func(_ context.Context, cond, arg string) error {
+		if cond != expectCond {
+			panic(fmt.Errorf("got condition %q want %q", cond, expectCond))
```
mattyw: maybe pass in c *gc.C and c.Assert?

bakery/checkers/checkers.go
rogpeppe: I think checkers.NewEmpty reads OK and goes well with checkers.New too.

bakery/checkers/checkers.go
rogpeppe: The panics are by design - the registration stuff is designed to happen up front (usually at init time), and if you're registering a duplicate name or haven't registered your URI, I think it's OK to panic, as that's entirely under your control.

If you're making a checker as a global variable (the way I want them to be able to be used), you'll end up panicking anyway if there's you get an error.

bakery/checkers/checkers_test.go
```diff
@@ -559,3 +427,113 @@ func (*CheckersSuite) TestOperationErrorCaveatTest(c *gc.C) {
 		c.Assert(test.caveat.Condition, gc.Matches, test.expectCondition)
 	}
 }
+
+func (*CheckersSuite) TestRegisterNilFuncPanics(c *gc.C) {
+	checker := checkers.New(nil)
```
mattyw, line 430: I don't think I've yet seen a test that checks the behaviour of chekers.New(nil)

bakery/checkers/declared.go
```diff
@@ -26,6 +40,7 @@ func DeclaredCaveat(key string, value string) Caveat {
 // wraps the provided third party caveat and requires
 // that the third party must add "declared" caveats for
 // all the named keys.
+// TODO namespaces in third party caveats?
```
mattyw: TODO (rog)

bakery/checkers/checkers.go
rogpeppe: I don't think it's needed here. Of the 221 implementations of sort.Interface in the the Go standard library and the packages under golang.org, I see exactly one that has a doc comment - the sort example.

bakery/service.go
rogpeppe: This type doesn't pull its weight any more. I've deleted it instead.

httpbakery/checkers.go
rogpeppe: done

bakery/checkers/declared.go
```diff
@@ -76,33 +77,42 @@ func (c Declared) Check(_, arg string) error {
 // different values, the information is omitted from the map.
 // When the caveats are later checked, this will cause the
 // check to fail.
-func InferDeclared(ms macaroon.Slice) Declared {
-	var conflicts []string
-	info := make(Declared)
+func InferDeclared(ns *Namespace, ms macaroon.Slice) map[string]string {
```
mattyw, line 80: The comment need updating as only conditions with Location "" are used

bakery/service.go
rogpeppe: deleted

bakery/checkers/declared.go
```diff
@@ -76,33 +77,42 @@ func (c Declared) Check(_, arg string) error {
...
-				conflicts = append(conflicts, key)
-				continue
-			}
-			info[key] = val
 		}
 	}
+	return InferDeclaredFromConditions(ns, conditions)
+}
+
+// InferDeclaredFromConditions is like InferDeclared except that
```
mattyw, line 92: as InferDeclared is basically the same as InferDeclaredFromConditions with filtering let's make sure the comments for these two functions make sense when considered as a pair

httpbakery/checkers.go
rogpeppe: done

bakery/checkers/namespace.go
```diff
@@ -0,0 +1,94 @@
...
+//
+// If ns is nil, it is treated as if it were empty.
+//
+// Resolve does not mutate ns and may be called concurrently
+// with other non-mutating Namespace methods.
+func (ns *Namespace) Resolve(uri string) (string, bool) {
+	if ns == nil {
+		return "", false
+	}
+	prefix, ok := ns.uriToPrefix[uri]
```
mattyw, line 44: return ns.uriToPrefix[uri]? (Does that work, is go able to infer you want to get from a map with bool from the return type of this function?

bakery/checkers/declared.go
```diff
@@ -76,33 +77,42 @@ func (c Declared) Check(_, arg string) error {
...
+// it is passed a set of first party caveat conditions rather than a set of macaroons.
+func InferDeclaredFromConditions(ns *Namespace, conds []string) map[string]string {
+	var conflicts []string
+	prefix, _ := ns.Resolve(StdNamespace)
```
mattyw, line 99: Worth a comment here that's it's safe to ignore the ok return here on the StdNamespace

bakery/checkers/namespace.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+	return ns
+}
+
+// EnsureResolved tries to resolve the given schema URI to a prefix and
+// returns the prefix and whether the resolution was successful. If the
+// URI hasn't been registered but a compatible version has, the
+// given URI is registered with the same prefix.
+func (ns *Namespace) EnsureResolved(uri string) (string, bool) {
+	// TODO compatibility
```
mattyw: TODO (rog)

bakery/checkers/namespace.go
```diff
@@ -0,0 +1,94 @@
...
+// it returns an error caveat.
+//
+// If ns.Namespace is empty or ns.Location is non-empty, it returns cav unchanged.
+//
+// If ns is nil, it is treated as if it were empty.
+//
+// ResolveCaveat does not mutate ns and may be called concurrently
+// with other non-mutating Namespace methods.
+func (ns *Namespace) ResolveCaveat(cav Caveat) Caveat {
+	// TODO If a namespace isn't registered, try to resolve it by
```
mattyw: TODO (rog)

bakery/checkers/namespace.go
```diff
@@ -0,0 +1,94 @@
...
+		cav.Condition = WithPrefix(prefix, cav.Condition)
+	}
+	cav.Namespace = ""
+	return cav
+}
+
+// WithPrefix returns the given string prefixed by the
+// given prefix. If the prefix is non-empty, a colon
+// is used to separate them.
+func WithPrefix(prefix, s string) string {
```
mattyw: checkers.WithPrefix is quite generic sounding, especially as afaik it's only to be used for conditions. If it's exported I'd like to be clearer when it should be used. checkers.CondWithPrefix or checkers.NamespacedCond

bakery/checkers/namespace.go
```diff
@@ -0,0 +1,94 @@
...
+	if prefix == "" {
+		return s
+	}
+	return prefix + ":" + s
+}
+
+// Register registers the given URI and associates it
+// with the given prefix. If the URI has already been registered,
+// this is no-op.
```
mattyw: a no-op

bakery/checkers/checkers.go
rogpeppe: 'cos it's a third party caveat, not a first party one. It will end up going probably, as I think third party caveats will change format quite a but.

bakerytest/bakerytest.go
```diff
@@ -291,6 +293,8 @@ func (d *InteractiveDischarger) wait(w http.ResponseWriter, r *http.Request) {
 				return cavs, nil
 			},
 		),
+		// TODO obtain the namespace from the client.
```
mattyw, line 296: TODO (rog)

bakery/checkers/checkers.go
rogpeppe: Yeah, I can see your reasons. I think (if it's OK) that I'll leave code movement alone for the moment - I think it might be worth moving the Checker code into its own file and having checkers.go for checker implementations only, but I've got a PR after this one that gives me conflicts on every change, so I'd prefer to wait for that if that's OK.

bakery/checkers/checkers.go
rogpeppe: I've added the Condition function.

bakery/checkers/checkers_test.go
rogpeppe: done

bakery/checkers/declared.go
rogpeppe: done

bakery/checkers/declared.go
rogpeppe: Conditions with Location=="" are the first party caveats. I don't see what the issue is.

bakery/checkers/declared.go
rogpeppe: I don't understand why they don't.

bakery/checkers/declared.go
rogpeppe: done

bakery/checkers/namespace.go
rogpeppe: Nope, that doesn't work.

bakery/checkers/namespace.go
rogpeppe: done

bakery/checkers/namespace.go
rogpeppe: done

bakery/checkers/namespace.go
rogpeppe: I renamed it to ConditionWithPrefix.
I don't think it'll be used often enough to justify the abbreviation.

bakery/checkers/namespace.go
rogpeppe: done

bakerytest/bakerytest.go
rogpeppe: done

bakery/checkers/checkers_test.go
rogpeppe: TestCheckers uses checkers.New(nil).

Are you thinking of some other kind of test?

bakery/checkers/checkers_test.go
rogpeppe: The dependency on bakery was annoying when refactoring because I couldn't run the checker tests without bakery compiling. It's always nicer to avoid cyclic dependencies even in tests when reasonable.

mattyw (approved):

PR #111, opened 2016-11-14
httpbakery: deprecate DoWithBody

rogpeppe:
DoWithBody is awkward and only exists because we
couldn't think of a way around the fact that http.NewRequest
hides the original type when used on seekable types
such as *bytes.Reader.

We work around that here by observing that we can actually
get the underlying reader out of the result of NopCloser,
meaning that we can just make the standard Do method work
as nature intended.

### DIFF
--- a/httpbakery/request.go
+++ b/httpbakery/request.go
@@ -1,0 +1,147 @@
+package httpbakery
+
+import (
+	"io"
+	"io/ioutil"
+	"net/http"
+	"reflect"
+	"sync"
+	"sync/atomic"
+
+	errgo "gopkg.in/errgo.v1"
+)
+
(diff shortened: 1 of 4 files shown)


httpbakery/request.go
```diff
@@ -0,0 +1,147 @@
...
+	"io/ioutil"
+	"net/http"
+	"reflect"
+	"sync"
+	"sync/atomic"
+
+	errgo "gopkg.in/errgo.v1"
+)
+
+// newRetrableRequest wraps an HTTP request so that it can
```
mhilton, line 14: newRetryableRequest

mhilton (approved):
LGTM

frankban (approved):
Looks good!

PR #113, opened 2016-11-15
httpbakery: retry requests more than once

rogpeppe:
Some servers can require more than one discharge (for example,
a server might require authentication to be able to see anything
at all, and then a third party discharge to be able to access an
individual entity).


mhilton (approved):
:+1:

alesstimec (approved):
LGTM, but i think the maxDischargeRetries might be too low... 3 is a reasonable number, i guess

PR #114, opened 2016-11-16
httpbakery/agent: simplify code by using httprequest

mhilton:


httpbakery/agent/agent.go
```diff
@@ -64,81 +63,43 @@ unmarshals to an httpbakery.Error.
...
-	if err != nil {
-		// This shouldn't happen as the agentLogin type has to be marshalable.
-		panic(errgo.Notef(err, "cannot marshal cookie"))
-	}
-	v := base64.StdEncoding.EncodeToString(b)
-	jar.SetCookies(u, []*http.Cookie{{
-		Name:  cookieName,
-		Value: v,
-	}})
+func (v *Visitor) AddAgent(u *url.URL, username string, key *bakery.KeyPair) {
```
rogpeppe: Doc comment please.

What are the intended semantics if we have two agents both with matching URLs for a request we're making?

httpbakery/agent/agent.go
```diff
@@ -64,81 +63,43 @@ unmarshals to an httpbakery.Error.
...
-// c.Key. c.VisitWebPage is set to VisitWebPage(c). The return is
-// non-nil only if c.Key is nil.
-func SetUpAuth(c *httpbakery.Client, u *url.URL, username string) error {
-	if c.Key == nil {
-		return errgo.New("cannot set-up authentication: client key not configured")
-	}
-	SetCookie(c.Jar, u, username, &c.Key.Public)
-	c.VisitWebPage = VisitWebPage(c)
-	return nil
+type Visitor struct {
```
rogpeppe: This needs a doc comment.

httpbakery/agent/agent.go
```diff
@@ -64,81 +63,43 @@ unmarshals to an httpbakery.Error.
...
 }
 
-// ErrNoAgentLoginCookie is the error returned when the expected
-// agent login cookie has not been found.
-var ErrNoAgentLoginCookie = errgo.New("no agent-login cookie found")
-
-// LoginCookie returns details of the agent login cookie
-// from the given request. If no agent-login cookie is found,
-// it returns an ErrNoAgentLoginCookie error.
-func LoginCookie(req *http.Request) (username string, key *bakery.PublicKey, err error) {
```
rogpeppe, line 87: Don't we still need this for the server side logic?

httpbakery/agent/agent_test.go
```diff
@@ -122,124 +122,118 @@ func (s *agentSuite) TestAgentLogin(c *gc.C) {
...
-
-	for i, test := range tests {
-		c.Logf("test %d: %s", i, test.about)
-
-		client := httpbakery.NewClient()
-		client.Key = key
-		test.setCookie(client, srvURL)
+var urlMatchTests = []struct {
+	about       string
+	url         *url.URL
```
rogpeppe: I think I'd specify these as strings - it'll be easier to read the test cases that way.

rogpeppe (commented):
Looks nice, thanks, and I think that being able to configure multiple agents for different endpoints. I have a few questions and suggestons.

httpbakery/agent/agent.go
mhilton: yes, looks like it.

httpbakery/agent/agent.go
mhilton: I intended first match wins.

rogpeppe: I wonder if most-specific-wins might be a better semantic - that's what cookies use.

httpbakery/agent/agent.go
```diff
@@ -64,81 +63,43 @@ unmarshals to an httpbakery.Error.
...
+type Visitor struct {
+	agents []agent
 }
 
-// SetCookie creates a cookie in jar which is suitable for performing agent
-// logins to u.
-//
-// If using SetUpAuth, it should not be necessary to use
-// this function.
-func SetCookie(jar http.CookieJar, u *url.URL, username string, pk *bakery.PublicKey) {
```
rogpeppe, line 127: I'm thinking that if we've still got LoginCookie, that this might still be worth exporting.

mhilton: I don't think it is by itself. I have exported it for local tests.

httpbakery/agent/agent.go
```diff
@@ -106,121 +123,116 @@ func LoginCookie(req *http.Request) (username string, key *bakery.PublicKey, err
...
+		return c == '/'
+	})
+}
+
+// pathCmp compares two slices of path parts and returns -1 if p1 comes
+// first logically, 1 if p1 comes second logically or 0 if they are
+// equal. The logical ordering used by this function compares each part
+// lexically, if these are equal up to the length of the shortest slice
+// then the longer slice comes first.
+func pathCmp(p1, p2 []string) int {
```
rogpeppe, line 184: Alternatively, you might consider this implementation from the cookiejar package:

```
// shouldSend determines whether e's cookie qualifies to be included in a
// request to host/path. It is the caller's responsibility to check if the
// cookie is expired.
func (e *entry) shouldSend(https bool, host, path string) bool {
    return e.domainMatch(host) && e.pathMatch(path) && (https || !e.Secure)
}

// domainMatch implements "domain-match" of RFC 6265 section 5.1.3.
func (e *entry) domainMatch(host string) bool {
    if e.Domain == host {
        return true
    }
    return !e.HostOnly && hasDotSuffix(host, e.Domain)
}

// pathMatch implements "path-match" according to RFC 6265 section 5.1.4.
func (e *entry) pathMatch(requestPath string) bool {
    if requestPath == e.Path {
        return true
    }
    if strings.HasPrefix(requestPath, e.Path) {
        if e.Path[len(e.Path)-1] == '/' {
            return true // The "/any/" matches "/any/path" case.
        } else if requestPath[len(e.Path)] == '/' {
            return true // The "/any" matches "/any/path" case.
        }
    }
    return false
}
```

httpbakery/agent/agent.go
```diff
@@ -106,121 +123,116 @@ func LoginCookie(req *http.Request) (username string, key *bakery.PublicKey, err
...
 		}
-		if err != nil {
-			return errgo.WithCausef(err, http.ErrNoCookie, "cannot perform agent login")
+	}
+	return append(agents, ag)
+}
+
+// pathParts splits a path up into an array of parts.
+func pathParts(path string) []string {
+	return strings.FieldsFunc(path, func(c rune) bool {
```
rogpeppe, line 174: Aren't trailing slashes meant to be significant? Maybe not, I'm not sure any more!

httpbakery/agent/agent.go
```diff
@@ -106,121 +123,116 @@ func LoginCookie(req *http.Request) (username string, key *bakery.PublicKey, err
...
+}
+
+// pathCmp compares two slices of path parts and returns -1 if p1 comes
+// first logically, 1 if p1 comes second logically or 0 if they are
+// equal. The logical ordering used by this function compares each part
+// lexically, if these are equal up to the length of the shortest slice
+// then the longer slice comes first.
+func pathCmp(p1, p2 []string) int {
+	for {
+		switch {
```
rogpeppe: I often use something like:

```
for len(p1) > 0 && len(p2) > 0 {
     if cmp := strings.Compare(p1[0], p2[0]); cmp != 0 {
          return cmp
     }
     p1, p2 = p1[1:], p2[1:]
 }
 if len(p1) > 0 {
    return -1
 }
 if len(p2) > 0 {
     return 1
 }
 return 0
```

which is a little simpler and I think more efficient.

httpbakery/agent/agent_test.go
```diff
@@ -149,97 +146,201 @@ func (s *agentSuite) TestLoginCookie(c *gc.C) {
...
 			continue
 		}
-		c.Assert(foundUser, gc.Equals, test.expectUser)
-		c.Assert(foundKey, gc.DeepEquals, test.expectKey)
+		c.Assert(username, gc.Equals, test.expectUser)
+		c.Assert(key, gc.DeepEquals, test.expectKey)
+	}
+}
+
+var pathCmpTests = []struct {
```
rogpeppe, line 285: I think I'd prefer to have a slightly higher level test than this (possibly in addition) which focused more on the external behaviour, because there's quite a bit of logic that happens before we ever get to pathCmp and it would be nice to be able to change the internal algorithm without changing the tests much.

rogpeppe (approved):
LGTM modulo a few comments and suggestions, thanks.

httpbakery/agent/agent.go
mhilton: done, It did take me a while to see how this works

httpbakery/agent/agent.go
mhilton: Only in as much as they are part of the matching algorithm, so a can be sent to be if a is a prefix of b and a ends with a /, or the unmatched part of be starts with a /.

httpbakery/agent/agent.go
mhilton: I don't think we should be overly tied to the domain and path matching algorithm of the cookie specification. hosts for example I think can just be exact matches. The path matching does work in the same way, but in a way I understand better.

httpbakery/agent/agent_test.go
mhilton: Done, I've added more cases to TestMultipleAgents

httpbakery/agent/agent.go
```diff
@@ -106,121 +123,111 @@ func LoginCookie(req *http.Request) (username string, key *bakery.PublicKey, err
...
+			continue
 		}
-		defer resp.Body.Close()
-		b, err := ioutil.ReadAll(resp.Body)
-		if err != nil {
-			logger.Errorf("cannot read response body: %s", err)
-			b = []byte{}
+		logger.Debugf("attempting login to %v using agent %s", url, agent.username)
+		c := &httprequest.Client{
+			Doer: &httpbakery.Client{
```
rogpeppe, line 213: I think this should probably make a shallow copy of client
and just set the key.

PR #115, opened 2016-11-18
bakery: implement checker type

rogpeppe:
This is the start of a replacement for the existing Service type. The
Service type does not pull its weight - it is only a thin wrapper around
functionality already exposed as functions/ It does not provide a useful
abstraction that makes it straightforward to build secure services with
macaroons. It also presupposes that there is a single authorization space
for macaroons - any macaroon can potentially authorize any operation,
limited only by first party caveats.

The new Checker type is a strict superset of the checking functionality
provided by Service. Every macaroon is associated one or more
*operations*, where an operation specifies (by name) an entity and an
action to take on that entity. The current Service behaviour is the same
as always creating macaroons with the LoginOp operation.

We bless our existing conventions for leveraging external identity
providers to authorize requests. A request may be authorized by virtue
of declared identity, but it also may be authorized by macaroons that
are associated with the required operations.

With the new Checker in place, it becomes possible to safely create
delegatable macaroons ("capabilities") that authorize a set of operations,
and it also becomes possible to arbitrarily combine these capabilities.

The Checker type does not itself create macaroons - this is left up to
the user of the Checker. There is another type that will be introduced
in a subsequent PR that will provide this functionality, at which point
the Service type can be removed.


bakery/authorizer.go
```diff
@@ -0,0 +1,134 @@
+package bakery
+
+import (
+	"golang.org/x/net/context"
+	"gopkg.in/errgo.v1"
+
+	"gopkg.in/macaroon-bakery.v2-unstable/bakery/checkers"
+)
+
+const Everyone = "everyone"
```
alesstimec: godoc, please..

bakery/authorizer.go
```diff
@@ -0,0 +1,134 @@
...
+	OpenAuthorizer openAuthorizer
+
+	// ClosedAuthorizer is an Authorizer implementation that will return ErrPermissionDenied
+	// on all authorization requests.
+	ClosedAuthorizer closedAuthorizer
+)
+
+type openAuthorizer struct{}
+
+func (openAuthorizer) Authorize(ctxt context.Context, id Identity, ops []Op) (allowed []bool, caveats []checkers.Caveat, err error) {
```
alesstimec, line 44: // Authorize implements the Authorizer interface.

bakery/authorizer.go
```diff
@@ -0,0 +1,134 @@
...
+var (
+	// OpenAuthorizer is an Authorizer implementation that will authorize all operations without question.
+	OpenAuthorizer openAuthorizer
+
+	// ClosedAuthorizer is an Authorizer implementation that will return ErrPermissionDenied
+	// on all authorization requests.
+	ClosedAuthorizer closedAuthorizer
+)
+
+type openAuthorizer struct{}
```
alesstimec, line 41: please add
var _ Authorizer = (_openAuthorizer)(nil)
var _ Authorizer = (_closedAuthorizer)(nil)

bakery/authorizer.go
```diff
@@ -0,0 +1,134 @@
...
+	allowed = make([]bool, len(ops))
+	for i := range allowed {
+		allowed[i] = true
+	}
+	return allowed, nil, nil
+}
+
+type closedAuthorizer struct{}
+
+func (closedAuthorizer) Authorize(ctxt context.Context, id Identity, ops []Op) (allowed []bool, caveats []checkers.Caveat, err error) {
```
alesstimec, line 55: // Authorize implements the Authorizer interface.

bakery/authorizer.go
```diff
@@ -0,0 +1,134 @@
...
+	return make([]bool, len(ops)), nil, nil
+}
+
+type authInfoKey struct{}
+
+func ContextWithAuthInfo(ctxt context.Context, authInfo *AuthInfo) context.Context {
```
alesstimec: godoc, please

bakery/authorizer.go
```diff
@@ -0,0 +1,134 @@
...
+	return context.WithValue(ctxt, authInfoKey{}, authInfo)
+}
+
+func AuthInfoFromContext(ctxt context.Context) *AuthInfo {
```
alesstimec: godoc, please

bakery/authorizer.go
```diff
@@ -0,0 +1,134 @@
...
+// See ACLAuthorizer for details.
+type ACLIdentity interface {
+	Identity
+
+	// Allow reports whether the user should be allowed to access
+	// any of the users or groups in the given ACL slice.
+	Allow(ctxt context.Context, acl []string) (bool, error)
+}
+
+func (a ACLAuthorizer) Authorize(ctxt context.Context, id Identity, ops []Op) (allowed []bool, caveats []checkers.Caveat, err error) {
```
alesstimec: // Authorize implements the Authorizer interface.

bakery/authorizer.go
```diff
@@ -0,0 +1,134 @@
...
+	// operations. It should return a slice with each element
+	// holding the ACL for the corresponding operation in the
+	// argument slice.
+	//
+	// If an entity cannot be found or the action is not recognised,
+	// GetACLs should return an empty ACL entry for that operation.
+	GetACLs func(ctxt context.Context, ops []Op) ([][]string, error)
+}
+
+// ACLIdentity may be implemented by Identity implementions
```
alesstimec: nitpick.. could you move this type definition to line 62? just to keep ACLAuthorizer implementation as one block

bakery/authorizer.go
```diff
@@ -0,0 +1,134 @@
...
+type ACLIdentity interface {
+	Identity
+
+	// Allow reports whether the user should be allowed to access
+	// any of the users or groups in the given ACL slice.
+	Allow(ctxt context.Context, acl []string) (bool, error)
+}
+
+func (a ACLAuthorizer) Authorize(ctxt context.Context, id Identity, ops []Op) (allowed []bool, caveats []checkers.Caveat, err error) {
+	if ops == nil {
```
alesstimec: shouldn't ops always be defined?

bakery/checker.go
```diff
@@ -0,0 +1,524 @@
...
+	macaroon "gopkg.in/macaroon.v2-unstable"
+
+	"gopkg.in/macaroon-bakery.v2-unstable/bakery/checkers"
+)
+
+// TODO think about a consistent approach to error reporting for macaroons.
+
+// TODO should we really pass in explicit expiry times on each call to Allow?
+
+var LoginOp = Op{
```
alesstimec, line 18: godoc, please

bakery/checker.go
```diff
@@ -0,0 +1,524 @@
...
+
+	// Entity holds the name of the entity to be authorized.
+	// Entity names should not contain spaces and should
+	// not start with the prefix "login" or "multi-" (conventionally,
+	// entity names will be prefixed with the entity type followed
+	// by a hyphen.
+	Entity string
+}
+
+var ErrPermissionDenied = errgo.New("permission denied")
```
alesstimec: godoc, please

bakery/checker.go
```diff
@@ -0,0 +1,524 @@
...
+
+type CheckerParams struct {
```
alesstimec, line 39: godoc, please

bakery/checker.go
```diff
@@ -0,0 +1,524 @@
...
+
+// allowAny is the internal version of AllowAny. Instead of returning an
+// authInfo struct, it returns a slice describing which operations have
+// been successfully authorized and a slice describing which macaroons
+// have been used in the authorization.
+func (a *AuthChecker) allowAny(ctxt context.Context, ops []Op) (authed, used []bool, err error) {
+	if err := a.init(ctxt); err != nil {
+		return nil, nil, errgo.Mask(err)
+	}
+	logger.Infof("after authorizer init, identity %#v", a.identity)
```
fabricematrat: debug left over ?

bakery/discharge_test.go
```diff
@@ -28,7 +28,7 @@ func (*DischargeSuite) TestDischargeAllNoDischarges(c *gc.C) {
 	rootKey := []byte("root key")
 	m, err := macaroon.New(rootKey, []byte("id0"), "loc0", macaroonCurrentVersion)
 	c.Assert(err, gc.IsNil)
-	ms, err := bakery.DischargeAll(m, noDischarge(c))
+	ms, err := bakery.DischargeAll(context.TODO(), m, noDischarge(c))
```
alesstimec: context.TODO()??

bakery/checker.go
```diff
@@ -0,0 +1,524 @@
...
+func (c *caveatSquasher) add(cond string) {
+	if c.add0(cond) {
+		c.conds = append(c.conds, cond)
+	}
+}
+
+func (c *caveatSquasher) add0(cond string) bool {
+	cond, args, err := checkers.ParseCaveat(cond)
+	if err != nil {
+		// Be safe - if we can't parse the caveat, just leave it there.
```
fabricematrat, line 428: maybe leave a log here ?

bakery/error.go
```diff
@@ -0,0 +1,65 @@
...
+	errgo "gopkg.in/errgo.v1"
+
+	"gopkg.in/macaroon-bakery.v2-unstable/bakery/checkers"
+)
+
+var (
+	// ErrNotFound is returned by Store.Get implementations
+	// to signal that an id has not been found.
+	ErrNotFound            = errgo.New("not found")
+	ErrCaveatResultUnknown = errgo.New("caveat result not known")
```
alesstimec: godoc, please

bakery/store.go
```diff
@@ -0,0 +1,144 @@
...
+	defer s.mu.Unlock()
+	if len(id) != 1 || id[0] != '0' || s.key == nil {
+		return nil, ErrNotFound
+	}
+	return s.key, nil
+}
+
+// RootKey implements Store.RootKey by
+//always returning the same root key.
+func (s *memRootKeyStore) RootKey(context.Context) (rootKey, id []byte, err error) {
```
alesstimec, line 54: why is context.Context being passed in?

bakery/identity.go
```diff
@@ -0,0 +1,44 @@
...
+	// Id returns the id of the user, which may be an
+	// opaque blob with no human meaning.
+	// An id is only considered to be unique
+	// with a given domain.
+	Id() string
+
+	// Domain holds the domain of the user. This
+	// will be empty if the user was authenticated
+	// directly with the identity provider.
+	Domain() string
```
fabricematrat, line 43: why do we need the domain ?

bakery/store.go
```diff
@@ -0,0 +1,144 @@
...
+	// caveat conditions that need to be checked.
+	//
+	// This method should not check first party caveats itself.
+	// TODO define some error type so we can distinguish store errors
+	// from bad ids and macaroon-not-found errors.
+	MacaroonOps(ctxt context.Context, ms macaroon.Slice) ([]Op, []string, error)
+}
+
+// OpsStore defines a persistent store for operation sets
+// stored by the Oven.
```
alesstimec: Oven??

bakery/store.go
```diff
@@ -0,0 +1,144 @@
...
+	// operations, for example by using a bloom filter.
+	GetOps(ctxt context.Context, key string) ([]Op, error)
+}
+
+type memOpsStore struct {
+	mu  sync.Mutex
+	ops map[string][]Op
+}
+
+func NewMemOpsStore() OpsStore {
```
alesstimec: godoc, please

bakery/checker.go
```diff
@@ -0,0 +1,524 @@
...
+	// IdentityClient.DeclaredIdentity.
+	Authorizer Authorizer
+
+	// IdentityClient is used for interactions with the external
+	// identity service used for authentication.
+	IdentityClient IdentityClient
+
+	// MacaroonOps is used to retrieve macaroon root keys
+	// and other associated information.
+	MacaroonOpStore MacaroonOpStore
```
fabricematrat, line 59: maybe embed it as only one ops on it ?

bakery/discharge_test.go
fabricematrat: euhh :) what is that ?

fabricematrat (approved):
LGTM but not fully understanding the whole :)

bakery/authorizer.go
```diff
@@ -0,0 +1,134 @@
...
+
+	"gopkg.in/macaroon-bakery.v2-unstable/bakery/checkers"
+)
+
+const Everyone = "everyone"
+
+// Authorizer is used to check whether a given user is allowed
+// to perform a set of operations.
+type Authorizer interface {
+	// Allow checks whether the given identity (which will be nil
```
mhilton: Does that method look like it's called Allow?

bakery/authorizer.go
```diff
@@ -0,0 +1,134 @@
...
+	// when there is no authenticated user) is allowed to perform
+	// the given operations. It should return an error only when
+	// some underlying database operation has failed, not when the
```
mhilton: I don't think this should necessarily talk about database operations. How about "It should return an error only when the authorization cannot be determined"?

bakery/authorizer.go
```diff
@@ -0,0 +1,134 @@
...
+	// user has been denied access.
+	//
+	// On success, each element of allowed holds whether the respective
+	// element of ops has been allowed, and caveats holds any additional
+	// third party caveats that apply.
+	Authorize(ctxt context.Context, id Identity, ops []Op) (allowed []bool, caveats []checkers.Caveat, err error)
```
mhilton, line 24: Should caveats be per operation as well?

bakery/authorizer.go
mhilton: if len(ops) == 0?

rogpeppe: Probably, but we're being defensive. From the point of view of the authorizer, we're asking whether the user can do nothing, and the answer is yes.

But I actually didn't intend to include this type in this PR. I'll remove it and add it later as it hasn't got decent tests yet anyway.

bakery/checker.go
rogpeppe: I don't think it's a good thing to use embedding unless you really want the properties it gives, which I don't in this case - I don't think CheckerParams should implement MacaroonOpStore.

bakery/checker.go
rogpeppe: I don't think we care - we'll get a log message if the caveat doesn't check out ok. And it's possible that we might have a valid caveat that we can't parse.

bakery/authorizer.go
```diff
@@ -0,0 +1,134 @@
...
+
+// ACLIdentity may be implemented by Identity implementions
+// to report group membership information.
+// See ACLAuthorizer for details.
+type ACLIdentity interface {
+	Identity
+
+	// Allow reports whether the user should be allowed to access
+	// any of the users or groups in the given ACL slice.
+	Allow(ctxt context.Context, acl []string) (bool, error)
```
mhilton: Should it be in the contract that Allow should always allow if the acl contains Everyone?

bakery/discharge_test.go
rogpeppe: context.TODO is for occasions when a call needs a context and you know you
should be providing a context taken from the caller, but you don't have
that context yet. It's not actually right to use here - I should be using BC (background
context) instead.

bakery/identity.go
rogpeppe: It's when you have multiple identity managers and one chains off the other. The Domain defines which identity manager did the discharge.

bakery/checker.go
```diff
@@ -0,0 +1,551 @@
...
+// Sometimes authorization will only be granted if a third party caveat
+// is discharged. This will happen when an IdentityClient or Authorizer
+// returns a third party caveat.
+//
+// When this happens, a DischargeRequiredError will be returned
+// containing the caveats and the operations required. The caller is
+// responsible for creating a macaroon with those caveats associated
+// with those operations and for passing that macaroon to the client to
+// discharge.
+type Checker struct {
```
mhilton, line 88: This doc comment is still too long for this type.

bakery/checker.go
```diff
@@ -0,0 +1,551 @@
...
+	FirstPartyCaveatChecker
+	p CheckerParams
+}
+
+// NewChecker returns a new Checker using the given parameters.
+// If p.CaveatChecker is nil, it will be initialized to checkers.New(nil).
```
mhilton: Why mention just one of the automatic things and not the others? probably just have it documented in the params type.

bakery/store.go
rogpeppe: Good question. Passing the context in here opens up the possibility of storing root keys in entity documents. The scenario I'm thinking of is this: we make a request to operate on an entity, which causes us to fetch the identity document. When retrieving the root key for the entity, we don't want to make another round trip to the database, so we attach the root key from the entity document to the context. Then RootKey can see that root key in the context and use it without needing to make a request to the database.

Essentially, passing the context in here allows any situation where the root key might be context-dependent.

bakery/store.go
rogpeppe: Ha, sneak preview. This type shouldn't be in this PR!

bakery/checker.go
```diff
@@ -0,0 +1,551 @@
...
+		//
+		// If the conditions fail, we won't use the macaroon for
+		// identity, but we can still potentially use it for its
+		// other operations if the conditions succeed for those.
+		declared, err := a.checkConditions(ctxt, LoginOp, conditions)
+		if err != nil {
+			a.initErrors = append(a.initErrors, errgo.Notef(err, "cannot authorize login macaroon"))
+			continue
+		}
+		if a.identity != nil {
```
mhilton, line 188: Above it says:

```
// The LoginOp operation is special - any macaroon associated with this
// operation is treated as a bearer of identity information. If two
// valid LoginOp macaroons are presented, only the first one will be
// used for identity.
```

But here it appears to be an error. Can you make these agree please?

bakery/authorizer.go
rogpeppe: Deleted, as it's not relevant to this PR.

bakery/authorizer.go
rogpeppe: done

bakery/authorizer.go
rogpeppe: done in the future but deleted for now as it's not relevant to this PR.

bakery/authorizer.go
rogpeppe: done in the future but deleted for now as it's not relevant to this PR.

bakery/checker.go
```diff
@@ -0,0 +1,551 @@
...
+
+// allowAny is the internal version of AllowAny. Instead of returning an
+// authInfo struct, it returns a slice describing which operations have
+// been successfully authorized and a slice describing which macaroons
+// have been used in the authorization.
+func (a *AuthChecker) allowAny(ctxt context.Context, ops []Op) (authed, used []bool, err error) {
+	if err := a.init(ctxt); err != nil {
+		return nil, nil, errgo.Mask(err)
+	}
+	logger.Infof("after authorizer init, identity %#v", a.identity)
```
mhilton: This looks like debugging to me

bakery/checker.go
```diff
@@ -0,0 +1,551 @@
...
+	// There are some unauthorized operations.
+	need := make([]Op, 0, len(ops)-numAuthed)
+	needIndex := make([]int, cap(need))
+	for i, ok := range authed {
+		if !ok {
+			needIndex[len(need)] = i
+			need = append(need, ops[i])
+		}
+	}
+	logger.Infof("operations needed after authz macaroons: %#v", need)
```
mhilton, line 322: This looks like debugging to me

bakery/checker.go
```diff
@@ -0,0 +1,551 @@
...
+
+	// Try to authorize the operations even even if we haven't got an authenticated user.
```
mhilton: s/even//

bakery/checker.go
```diff
@@ -0,0 +1,551 @@
...
+	oks, caveats, err := a.p.Authorizer.Authorize(ctxt, a.identity, need)
+	if err != nil {
+		return authed, used, errgo.Notef(err, "cannot check permissions")
+	}
+	if len(oks) != len(need) {
+		return authed, used, errgo.Newf("unexpected slice length returned from Allow (got %d; want %d)", len(oks), len(need))
```
mhilton: I wonder whether this sort of error, which presumably represents a bug in the surrounding application, should have a particular cause?

bakery/checker.go
```diff
@@ -0,0 +1,551 @@
...
+			authed[needIndex[i]] = true
+		} else {
+			stillNeed = append(stillNeed, ops[needIndex[i]])
+		}
+	}
+	if len(stillNeed) == 0 && len(caveats) == 0 {
+		// No more ops need to be authenticated and no caveats to be discharged.
+		return authed, used, nil
+	}
+	logger.Infof("operations still needed after auth check: %#v", stillNeed)
```
mhilton, line 342: more debugging

bakery/checker.go
```diff
@@ -0,0 +1,551 @@
...
+// If ops contains LoginOp, the user must have been authenticated with a
+// macaroon associated with the single operation LoginOp only.
+func (a *AuthChecker) AllowCapability(ctxt context.Context, ops ...Op) ([]string, error) {
+	nops := 0
+	for _, op := range ops {
+		if op != LoginOp {
+			nops++
+		}
+	}
+	if nops == 0 {
```
mhilton, line 388: Is it important this is an error? could we not just process it anyway?

bakery/checker.go
```diff
@@ -0,0 +1,551 @@
...
+		return nil, errgo.Newf("no non-login operations required in capability")
+	}
+	_, used, err := a.allowAny(ctxt, ops)
+	if err != nil {
+		logger.Infof("allowAny returned used %v; err %v", used, err)
```
mhilton: debug?

bakery/checker.go
```diff
@@ -0,0 +1,551 @@
...
+			squasher.add(cond)
+		}
+	}
+	return squasher.final(), nil
+}
+
+// caveatSquasher rationalizes first party caveats created for a capability
+// by:
+//	- including only the earliest time-before caveat.
+//	- excluding allow and deny caveats (operations are checked by
```
mhilton, line 410: This worries me slightly. This is the only place in bakery itself where the contents of caveat strings are understood. Perhaps this stuff should be in the checkers package somehow.

bakery/checker.go
```diff
@@ -0,0 +1,551 @@
...
+func (c *caveatSquasher) final() []string {
+	if !c.expiry.IsZero() {
+		c.conds = append(c.conds, checkers.TimeBeforeCaveat(c.expiry).Condition)
+	}
+	if len(c.conds) == 0 {
+		return nil
+	}
+	// Make deterministic and eliminate duplicates.
+	sort.Strings(c.conds)
+	logger.Infof("after sorting ops: %v", c.conds)
```
mhilton: debug

bakery/checker.go
```diff
@@ -0,0 +1,551 @@
...
+	j := 1
+	for _, cond := range c.conds[1:] {
+		if cond != prev {
+			c.conds[j] = cond
+			prev = cond
+			j++
+		}
+	}
+	c.conds = c.conds[:j]
+	logger.Infof("after deduplication: %q", c.conds)
```
mhilton: debug

bakery/checker.go
```diff
@@ -0,0 +1,551 @@
...
+	return c.conds
+}
+
+func (a *AuthChecker) checkConditions(ctxt context.Context, op Op, conds []string) (map[string]string, error) {
+	logger.Infof("checking conditions %q", conds)
```
mhilton: debug

bakery/checker.go
rogpeppe: turned it into Debugf, as it may still be useful.

mhilton (commented):
Comments for the first few files.

bakery/checker_test.go
```diff
@@ -0,0 +1,933 @@
...
+	"golang.org/x/net/context"
+	gc "gopkg.in/check.v1"
+	errgo "gopkg.in/errgo.v1"
+	"gopkg.in/macaroon.v2-unstable"
+
+	"gopkg.in/macaroon-bakery.v2-unstable/bakery"
+	"gopkg.in/macaroon-bakery.v2-unstable/bakery/checkers"
+)
+
+var logger = loggo.GetLogger("bakery.auth_test")
```
mhilton: bakery.test / bakery_test ?

bakery/checker_test.go
```diff
@@ -0,0 +1,933 @@
...
+
+const Everyone = "everyone"
```
mhilton, line 21: Why not use bakery.Everyone?

bakery/checker_test.go
```diff
@@ -0,0 +1,933 @@
...
+// or the slice contains "everyone", authorization is granted.
+type opAuthorizer map[bakery.Op][]string
+
+func (auth opAuthorizer) Authorize(ctxt context.Context, id bakery.Identity, ops []bakery.Op) (allowed []bool, caveats []checkers.Caveat, err error) {
+	return authorizerFunc(auth.authorize).Authorize(ctxt, id, ops)
+}
+
+func (auth opAuthorizer) authorize(id string, op bakery.Op) (bool, []checkers.Caveat, error) {
+	for _, allowedId := range auth[op] {
+		if allowedId == "everyone" || id == allowedId {
```
mhilton: You've got two constants you can use for everyone, and you still put in the string!

bakery/authorizer.go
rogpeppe: If you squint a lot.

bakery/authorizer.go
rogpeppe: Good suggestion, done.

bakery/authorizer.go
rogpeppe: I decided not, on the basis that we probably don't want to return many discharge-required errors but just one containing all the caveats. Can you think of a case where we'd want to know which of the allowed operations required which caveats?

bakery/checker.go
rogpeppe: done, and moved to the errors.go file.

bakery/checker.go
rogpeppe: done

bakery/checker.go
rogpeppe: Moved it to doc.go.

bakery/checker.go
rogpeppe: Yeah, I did that just before I saw this comment :)

bakery/error.go
```diff
@@ -0,0 +1,65 @@
...
+func (e *DischargeRequiredError) Error() string {
+	return "macaroon discharge required: " + e.Message
+}
+
+func isDischargeRequiredError(err error) bool {
+	_, ok := err.(*DischargeRequiredError)
+	return ok
+}
+
+// VerificationError is ... TODO
```
mhilton: This should probably be done soon, before you forget

bakery/checker.go
rogpeppe: Good point, done.

bakery/checker.go
rogpeppe: After another look, yes, it's redundant. Removed.

bakery/checker.go
rogpeppe: I changed it to Debugf.

bakery/checker.go

bakery/store.go
```diff
@@ -0,0 +1,78 @@
...
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if len(id) != 1 || id[0] != '0' || s.key == nil {
+		return nil, ErrNotFound
+	}
+	return s.key, nil
+}
+
+// RootKey implements Store.RootKey by
+//always returning the same root key.
```
mhilton: you need a space after the // here

mhilton (approved):
Great stuff, thanks

bakery/checker.go
rogpeppe: I dunno. I don't want callers checking for it particularly. It's almost a panic.

I've changed the definition of Authorize so that this check isn't necessary
(additional elements are treated as false).

bakery/checker.go
rogpeppe: changed to Debugf

bakery/checker.go
rogpeppe: It's actually quite important, as capabilities should not in general be granted for the LoginOp operation, which is intended to be reserved for actual client authentication. If you're not granting a capability for LoginOp and you've got no other operations, then you have a problem - we don't want to issue macaroons for no operations at all. Hence the error.

I've updated the doc comment to try to say some of this.

bakery/checker.go
rogpeppe: removed

bakery/authorizer.go
mhilton: I can't think of one, I was just looking at the consistency. I'm happy to trust your judgment here.

bakery/checker.go
rogpeppe: I _think_ it's reasonable for the checker to enforce checking of the operations caveats, as it's in the best place to know when to add the operation context when checking for a particular operation.

I also think it's reasonable for the bakery to add the time-before caveat when it mints a macaroon with a given expiry time, and it parses declared caveats when working out declarations for the identity client.

One possibility might be to move the time-before and operations checkers out of the checkers package into bakery, I guess.

bakery/checker.go
rogpeppe: removed

bakery/checker.go
rogpeppe: removed.

bakery/checker.go
rogpeppe: removed

bakery/store.go
rogpeppe: done

bakery/error.go
rogpeppe: done

bakery/checker_test.go
rogpeppe: done

bakery/checker_test.go
rogpeppe: 'cos it's been removed pending the next PR.

bakery/checker_test.go
rogpeppe: done

bakery/checker.go
```diff
@@ -0,0 +1,479 @@
...
+	"time"
+
+	"golang.org/x/net/context"
+	errgo "gopkg.in/errgo.v1"
+	macaroon "gopkg.in/macaroon.v2-unstable"
+
+	"gopkg.in/macaroon-bakery.v2-unstable/bakery/checkers"
+)
+
+var LoginOp = Op{
```
alesstimec, line 18: godoc, please

bakery/checker.go
```diff
@@ -0,0 +1,479 @@
...
+	// or "delete". It is up to the service using a checker to define
+	// a set of operations and keep them consistent over time.
+	Action string
+
+	// Entity holds the name of the entity to be authorized.
+	// Entity names should not contain spaces and should
+	// not start with the prefix "login" or "multi-" (conventionally,
+	// entity names will be prefixed with the entity type followed
+	// by a hyphen.
+	Entity string
```
alesstimec, line 35: could we have a Validate method that would check the constrains on the Entity string?

alesstimec: do we need one?

bakery/checker.go
```diff
@@ -0,0 +1,479 @@
...
+	for i, ms := range a.macaroons {
+		ops, conditions, err := a.p.MacaroonOpStore.MacaroonOps(ctxt, ms)
+		if err != nil {
+			if !isVerificationError(err) {
+				return errgo.Notef(err, "cannot retrieve macaroon")
+			}
+			a.initErrors = append(a.initErrors, errgo.Mask(err))
+			continue
+		}
+		logger.Infof("checking macaroon %d; ops %q, conditions %q", i, ops, conditions)
```
alesstimec: Debugf, perhaps?

bakery/checker_test.go
```diff
@@ -0,0 +1,945 @@
...
+	errgo "gopkg.in/errgo.v1"
+	"gopkg.in/macaroon.v2-unstable"
+
+	"gopkg.in/macaroon-bakery.v2-unstable/bakery"
+	"gopkg.in/macaroon-bakery.v2-unstable/bakery/checkers"
+)
+
+var logger = loggo.GetLogger("bakery.bakery_test")
+
+const Everyone = "everyone"
```
alesstimec, line 21: does not need to be exported

bakery/common_test.go
```diff
@@ -0,0 +1,149 @@
...
+	gc "gopkg.in/check.v1"
+	"gopkg.in/macaroon.v2-unstable"
+
+	"gopkg.in/macaroon-bakery.v2-unstable/bakery"
+	"gopkg.in/macaroon-bakery.v2-unstable/bakery/checkers"
+)
+
+// BC holds the testing background context - its associated time when checking
+// time-before caveats will always be the value of epoch.
+var BC = checkers.ContextWithClock(context.Background(), stoppedClock{epoch})
```
alesstimec: does not need to be exported

bakery/service.go
```diff
@@ -387,13 +376,16 @@ func (svc *Service) AddCaveat(m *macaroon.Macaroon, cav checkers.Caveat) error {
 // resulting third-party caveat will encode the condition "true"
 // encrypted with that public key. See LocalThirdPartyCaveat
 // for a way of creating such caveats.
-func AddCaveat(key *KeyPair, loc ThirdPartyLocator, m *macaroon.Macaroon, cav checkers.Caveat, ns *checkers.Namespace) error {
+func AddCaveat(ctxt context.Context, key *KeyPair, loc ThirdPartyLocator, m *macaroon.Macaroon, cav checkers.Caveat, ns *checkers.Namespace) error {
 	if cav.Location == "" {
```
alesstimec, line 380: check 
if m == nil 
if ns == nil

bakery/store.go
```diff
@@ -0,0 +1,81 @@
...
+// NewMemRootKeyStore returns an implementation of
+// Store that generates a single key and always
+// returns that from RootKey. The same id ("0") is always
+// used.
+func NewMemRootKeyStore() RootKeyStore {
+	return new(memRootKeyStore)
+}
+
+type memRootKeyStore struct {
+	mu  sync.Mutex
```
alesstimec, line 38: could use sync.RWMutex..  Get method reads the key and could use RLock and RUnlock..  or sync/atomic Value

alesstimec (approved):
A few extra checks, a few nitpicks, but modulo that: LGTM. nice work!

bakery/checker.go
rogpeppe: done

bakery/checker.go
rogpeppe: I don't think we need one. It's really up to the caller to make sure that their entities are valid for their use case.

bakery/checker.go

bakery/checker_test.go
rogpeppe: leaving it exported, 'cos it'll be bakery.Everyone in the next PR and I want to make the global replace easy :)

bakery/common_test.go
rogpeppe: Since it's in a test, it doesn't matter and I on balance prefer the visual distinctiveness of BC vs bc.
I think I'm swaying away from BC towards testContext actually. I'll change everything to use that.

bakery/service.go
rogpeppe: It's OK to use a nil namespace, and I think that if you're adding a caveat to a nil macaroon you deserve the panic you'll get.

bakery/store.go
rogpeppe: The logic is simpler with Mutex and this code isn't really intended for production, as real services will need to store root keys persistently.
If we ever benchmark this code as a point of contention, we'll do as you say.

PR #116, opened 2016-11-18
checkers: implement serialization for namespaces

rogpeppe:
This will allow us to pass namespaces around with macaroons
so that clients can know what first party caveats can be added.

A namespace is a many-to-one mapping of schema URI to
prefix which we represent as a space separated set
of URI:prefix pairs.

### DIFF
--- a/bakery/checkers/namespace.go
+++ b/bakery/checkers/namespace.go
@@ -1,5 +1,14 @@
 package checkers
 
+import (
+	"sort"
+	"strings"
+	"unicode"
+	"unicode/utf8"
+
+	errgo "gopkg.in/errgo.v1"
+)
+
 // Namespace holds maps from schema URIs to the
 // prefixes that are used to encode them in first party
 // caveats. Several different URIs may map to the same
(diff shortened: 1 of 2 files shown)


bakery/checkers/namespace.go
```diff
@@ -10,17 +19,80 @@ type Namespace struct {
...
+	uriToPrefix := make(map[string]string)
+	elems := strings.Fields(string(data))
+	for _, elem := range elems {
+		i := strings.LastIndex(elem, ":")
+		if i == -1 {
+			return errgo.Newf("no colon in namespace field %q", elem)
+		}
+		uri, prefix := elem[0:i], elem[i+1:]
+		if !IsValidSchemaURI(uri) {
+			// Currently this can't happen because the only invalid URIs
```
mhilton, line 80: Or are empty which could easily happen.

bakery/checkers/namespace.go
```diff
@@ -77,18 +149,43 @@ func (ns *Namespace) ResolveCaveat(cav Caveat) Caveat {
...
+
+func IsValidSchemaURI(uri string) bool {
+	// TODO more stringent requirements?
+	return len(uri) > 0 &&
+		utf8.ValidString(uri) &&
+		strings.IndexFunc(uri, invalidSchemaRune) == -1
+}
+
+func invalidPrefixRune(r rune) bool {
+	return r == ' ' || r == ':' || unicode.IsSpace(r)
```
mhilton, line 186: why have r == ' ' in this one and not the one above?

mhilton (approved):
:+1:

bakery/checkers/namespace.go
rogpeppe: 'cos I forgot to remove it.

bakery/checkers/namespace.go
rogpeppe: indeed, and I even covered it with a test case already.

bakery/checkers/namespace.go
```diff
@@ -77,18 +149,43 @@ func (ns *Namespace) ResolveCaveat(cav Caveat) Caveat {
...
+		return condition
 	}
-	return prefix + ":" + s
+	return prefix + ":" + condition
 }
 
 // Register registers the given URI and associates it
 // with the given prefix. If the URI has already been registered,
 // this is a no-op.
 func (ns *Namespace) Register(uri, prefix string) {
```
alesstimec, line 162: does this have to be exported

alesstimec (approved):

bakery/checkers/namespace.go
rogpeppe: Yes, it's important that it's exported - if you're defining some schema for first party caveats, you'll need to register the schema with the namespace.

PR #118, opened 2016-11-22
httpbakery: Add context to to client methods.

mhilton:
Add a DoWithContext method and use the context with all HTTP
requests. Also add context to DischargeAll and HandleError methods.

### DIFF
--- a/httpbakery/client.go
+++ b/httpbakery/client.go
@@ -14,6 +14,7 @@ import (
 	"github.com/juju/httprequest"
 	"github.com/juju/loggo"
 	"golang.org/x/net/context"
+	"golang.org/x/net/context/ctxhttp"
 	"golang.org/x/net/publicsuffix"
 	"gopkg.in/errgo.v1"
 	"gopkg.in/macaroon.v2-unstable"
@@ -146,7 +147,7 @@ type Client struct {
 type DischargeAcquirer interface {
 	// AcquireDischarge should return a discharge macaroon for the given third
 	// party caveat.
-	AcquireDischarge(cav macaroon.Caveat) (*macaroon.Macaroon, error)
+	AcquireDischarge(ctx context.Context, cav macaroon.Caveat) (*macaroon.Macaroon, error)
 }
 
 // NewClient returns a new Client containing an HTTP client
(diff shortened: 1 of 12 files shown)


mattyw (approved):

httpbakery/client.go
```diff
@@ -198,20 +180,30 @@ func (c *Client) dischargeAcquirer() DischargeAcquirer {
...
+// If the request fails with a discharge-required error, any required
+// discharge macaroons will be acquired, and the request will be repeated
+// with those attached.
+//
+// If the required discharges were refused by a third party, an error
+// with a *DischargeError cause will be returned.
+//
+// If interaction is required by the user, the client's WebPageVisitor,
+// or VisitWebPage function will be used to perform interaction. An error
+// with a *InteractionError cause will be returnedif this interaction
```
rogpeppe: returned if

httpbakery/client.go
```diff
@@ -518,6 +510,22 @@ func (c *Client) interact(location, visitURLStr, waitURLStr string) (*macaroon.M
 	return resp.Macaroon, nil
 }
 
+// relativeURL returns newPath relative to an original URL.
```
rogpeppe, line 513: Thanks for moving this.

httpbakery/client_test.go
```diff
@@ -238,7 +238,7 @@ func assertDischargeServerDischargesConditionForVersion(c *gc.C, cond string, ve
 	}, nil)
 	c.Assert(err, gc.IsNil)
```
rogpeppe, line 239: I wonder if we should test whether the context actually gets through to the parts we intend it to.

httpbakery/client.go
```diff
@@ -198,20 +180,30 @@ func (c *Client) dischargeAcquirer() DischargeAcquirer {
...
+// with a *DischargeError cause will be returned.
+//
+// If interaction is required by the user, the client's WebPageVisitor,
+// or VisitWebPage function will be used to perform interaction. An error
+// with a *InteractionError cause will be returnedif this interaction
+// fails. See WebBrowserVisitor for a possible implementation of
+// WebPageVisitor.
+//
+// DoWithContext may add headers to req.Header.
+func (c *Client) DoWithContext(ctx context.Context, req *http.Request) (*http.Response, error) {
```
rogpeppe, line 205: I'm not sure about having another Do variant again now we've eliminated DoWithBody.. For example, what about DoWithContextAndCustomError?

How about just adding a context to Do and DoWithCustomError and don't bother to derive the context from the request?
At some point when we can use Go 1.7 we can break the API and take the context argument out and derive it from the request.

For now, I think all our pre-1.7 programs will need to pass in the request explicitly anyway.

rogpeppe (changes requested):
LGTM but I have a concern about the proliferation of Do methods.

httpbakery/client.go
mhilton: done

httpbakery/client.go
mhilton: As discussed this is the interface we are using with httprequest

httpbakery/client.go
mhilton: you're welcome

rogpeppe (approved):

PR #119, opened 2016-11-23
bakery: add authorizer helper types

rogpeppe:
These are designed to help writing tests and to support
our usual ACL model.

We also update the package docs a bit.

### DIFF
--- a/bakery/authorizer.go
+++ b/bakery/authorizer.go
@@ -2,6 +2,7 @@ package bakery
 
 import (
 	"golang.org/x/net/context"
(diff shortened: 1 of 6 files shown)


bakery/authorizer.go
```diff
@@ -36,6 +36,8 @@ var (
 var (
```
mhilton, line 36: shouldn't this stuff be in a test file somewhere?

mhilton (approved):
:+1:

alesstimec (approved):
LGTM

PR #120, opened 2016-11-28
httpbakery/agent: export Agent type

rogpeppe:
We export the Agent type in preparation for providing
serialization support for Visitor. We also
make Agent the internal representation too
and use a slightly more efficient, allocation-free, lookup
algorithm.

### DIFF
--- a/httpbakery/agent/agent.go
+++ b/httpbakery/agent/agent.go
@@ -12,6 +12,7 @@ import (
 	"errors"
 	"net/http"
 	"net/url"
+	"sort"
 	"strings"
 
 	"github.com/juju/httprequest"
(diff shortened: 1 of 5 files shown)


httpbakery/agent/agent.go
```diff
@@ -123,11 +124,21 @@ func LoginCookie(req *http.Request) (username string, key *bakery.PublicKey, err
...
+}
+
+// agent represents an agent that can be used for agent authentication.
+type Agent struct {
+	// URL holds the URL associated with the agent.
+	URL string
+	// Username holds the username to use for the agent.
+	Username string
+	// Key holds the agent's private key pair.
+	Key *bakery.KeyPair
```
mhilton, line 141: Using a KeyPair representation will be awkward if you want to use the same type for serializing as an agent-cookie.

mhilton: Also with existing serializations already in production (well staging)

mhilton (approved):
:+1:

httpbakery/agent/agent.go
rogpeppe: I don't think we need to use the same type for the agent cookie. I see them as two different things, even though one can be derived from the other.

httpbakery/agent/agent.go
```diff
@@ -136,98 +147,151 @@ type Visitor struct {
...
+
+// AddAgent adds an agent to the visitor. The agent information will be
+// used when sending discharge requests to all URLs under the given URL.
+// If more than one agent matches a target URL then the agent with the
+// most specific matching URL will be used. Longer paths are counted as
+// more specific than shorter paths.
+//
+// Unlike HTTP cookies, a trailing slash is not significant, so for
+// example, if an agent is registered with the URL
+// http://kremvax.com/foo, its information will be sent to
```
jrwren: Maybe use example.com here instead of kremvax.com.

jrwren (approved):

httpbakery/agent/agent.go
rogpeppe: done

PR #123, opened 2016-12-08
httprequest/agent: use standard bakerytest discharger

mhilton:
Update the agent tests to use the InteractiveDischarger from bakery test. Removing the need for a lot of duplicated code.


rogpeppe:
merged in #124 .

PR #137, opened 2017-01-05
httpbakery: allow empty origin headers when checking caveats

rogpeppe:
Note: this is landing in a feature branch so that it can be used
by idm without the need to update it to the latest bakery API.


frankban (approved):
Looks good, thanks for the quick change!

jrwren (approved):

PR #139, opened 2017-01-12
bakery/checkers: simplify declarations

rogpeppe:
As implied by the proposal document, we make declaration caveats simpler
and hopefully a little less error prone.

In the new scheme, instead of being able to declare an arbitrary
set of attribute/value pairs, each third party will in general use
only a single caveat to declare its information.

This should be less vulnerable to security flaws due to
attributes being added, because all attributes should be declared
within the same caveat.

For backward compatibility, we still support a limited form
of the old scheme, sufficient to declare a single attribute.
This should be sufficient for our existing identity manager
which always uses "declared username" and declares no other
attributes.

The need-declared mechanism is due to change too - we
still support the old scheme, but will implement the new scheme
in a subsequent PR.

### DIFF
--- a/bakery/checker.go
+++ b/bakery/checker.go
@@ -128,18 +128,27 @@ func (c *Checker) Auth(mss ...macaroon.Slice) *AuthChecker {
 type AuthChecker struct {
 	// Checker is used to check first party caveats.
 	*Checker
+
+	// macaroons holds the macaroons providing authorization.
 	macaroons []macaroon.Slice
+
+	// initOnce guards all the fields below it.
+	initOnce   sync.Once
+	initError  error
+	initErrors []error
 	// conditions holds the first party caveat conditions
 	// that apply to each of the above macaroons.
-	conditions      [][]string
-	initOnce        sync.Once
-	initError       error
-	initErrors      []error
-	identity        Identity
+	conditions [][]string
+	// identity holds the authenticated idenity.
+	identity Identity
+	// When identity is nil, identityCaveats may hold the third party caveats that
+	// can be used to make a macaroon that will prove the identity.
 	identityCaveats []checkers.Caveat
 	// authIndexes holds for each potentially authorized operation
 	// the indexes of the macaroons that authorize it.
 	authIndexes map[Op][]int
+	// declarationCondition holds the resolved identity declaration condition
+	declarationCondition string
 }
 
 func (a *AuthChecker) init(ctxt context.Context) error {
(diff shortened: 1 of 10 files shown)


bakery/service.go
```diff
@@ -154,8 +144,24 @@ func Discharge(ctx context.Context, p DischargeParams) (*Macaroon, error) {
...
+	if cavInfo.Version < Version3 {
+		return checkThirdPartyCaveatPreVersion3(ctx, cavInfo, checker)
+	}
+	// TODO(rog) new-style need-declared checking
+	return checker.CheckThirdPartyCaveat(ctx, cavInfo)
+}
+
+func checkThirdPartyCaveatPreVersion3(ctx context.Context, cavInfo *ThirdPartyCaveatInfo, checker ThirdPartyCaveatChecker) ([]checkers.Caveat, error) {
+	// Note that we don't check the error - we allow the
+	// third party checker to see even caveats that we can't
```
mhilton: The word even in this sentence seems to confuse things.

bakery/service.go
```diff
@@ -229,8 +241,8 @@ type ThirdPartyCaveatInfo struct {
 	// RootKey holds the secret root key encoded by the caveat.
 	RootKey []byte
 
-	// CaveatId holds the full encoded caveat id from which all
-	// the other fields are derived.
+	// CaveatId holds the full encoded caveat from which all
```
mhilton: Caveat

mhilton (approved):
LGTM

bakery/checker.go
```diff
@@ -434,6 +455,8 @@ func (c *caveatSquasher) add(cond string) {
 	}
 }
 
+// add0 adds the caveat to the squasher. It reports whether
+// the caveat should actually been added.
```
alesstimec: s/been/be/?

bakery/identity.go
```diff
@@ -53,9 +60,13 @@ func (noIdentities) IdentityFromContext(ctxt context.Context) (Identity, []check
 	return nil, nil, nil
 }
 
+func (noIdentities) DeclarationCaveat() checkers.Caveat {
```
alesstimec, line 65: godoc, please

alesstimec (approved):
LGTM modulo  few nitpicks

bakery/service.go
rogpeppe: done

bakery/service.go
rogpeppe: done

bakery/checker.go

bakery/identity.go
rogpeppe: done

PR #140, opened 2017-01-12
bakery: use []byte not string for third party caveat condition

rogpeppe:
This paves the way for encoding more information compactly
inside a third party caveat.


mhilton (approved):
LGTM

alesstimec (approved):

PR #141, opened 2017-01-12
use standardized spelling for "ctx"

rogpeppe:
Everywhere in the stdlib and in our other repositories uses
ctx, so let's use it here.

This is an entirely mechanical change.


bac (approved):

mhilton (approved):
LGTM

PR #142, opened 2017-01-12
bakery: rename service.go to discharge.go

rogpeppe:
The name "service.go" was appropriate when it contained the
Service type, but as that type no longer exists, it seems wrong.

As we've already got a discharge.go file, rename that to
dischargeall.go to reflect that file's function more specifically.


rogpeppe:
merging as trivial.

bac (approved):

PR #143, opened 2017-01-12
bakery: rename discharge.go to dischargeall.go

rogpeppe:
This prepares for renaming service.go to discharge.go
hopefully in a way so that git can track the file renames


rogpeppe:
Landing as trivial

PR #144, opened 2017-01-12
bakery: define new encoding format for third party caveats

rogpeppe:
Instead of a string, we will use a protobuf encoding
that will include the need-declared information as
a field and the actual condition body as a binary
byte slice.

Note that the .pb.go files at automatically generated.


alesstimec (approved):

bakery/authorizer_test.go
```diff
@@ -27,14 +27,12 @@ func (*authorizerSuite) TestAuthorizerFunc(c *gc.C) {
 		case "b":
 			return true, nil, nil
 		case "c":
-			return true, []checkers.Caveat{{
-				Location:  "somewhere",
-				Condition: "c",
+			return true, []checkers.Caveat{{Location: "somewhere",
```
mhilton, line 30: This seems an odd formatting, why change it.

bakery/authorizer_test.go
```diff
@@ -27,14 +27,12 @@ func (*authorizerSuite) TestAuthorizerFunc(c *gc.C) {
...
+				ThirdPartyCondition: []byte("c"),
 			}}, nil
 		case "d":
-			return true, []checkers.Caveat{{
-				Location:  "somewhere",
-				Condition: "d",
+			return true, []checkers.Caveat{{Location: "somewhere",
```
mhilton, line 34: and this one

bakery/checkers/checkers.go
```diff
@@ -169,18 +169,39 @@ func checkError(ctx context.Context, _, arg string) error {
...
 
 // Caveat represents a condition that must be true for a check to
-// complete successfully. If Location is non-empty, the caveat must be
-// discharged by a third party at the given location.
-// The Namespace field holds the namespace URI of the
-// condition - if it is non-empty, it will be converted to
-// a namespace prefix before adding to the macaroon.
+// complete successfully. It can represent either a first-party
+// or a third-party caveat.
 type Caveat struct {
```
mhilton, line 174: It seems to me that there is something wrong with having a type where all the fields a mutually exclusive (no field is shared between first and third party caveats). Perhaps this should be two types with some kind of common interface?

mhilton (approved):
LGTM with a few thoughts

httpbakery/client_test.go
```diff
@@ -1271,10 +1271,10 @@ func newDischargeRequiredError(hp serverHandlerParams, checkErr error, req *http
 	}
 	var caveats []checkers.Caveat
 	if hp.authLocation != "" {
-		caveats = []checkers.Caveat{{
-			Location:  hp.authLocation,
-			Condition: "is-ok",
+		caveats = []checkers.Caveat{{Location: hp.authLocation,
```
mhilton, line 1274: Fromatting again

httpbakery/client_test.go
```diff
@@ -423,10 +423,10 @@ func (s *ClientSuite) serverRequiringMultipleDischarges(n int, discharger *baker
 			w.Write([]byte("ok"))
 			return
 		}
-		caveats := []checkers.Caveat{{
-			Location:  discharger.Location(),
-			Condition: "is-ok",
+		caveats := []checkers.Caveat{{Location: discharger.Location(),
```
mhilton, line 426: And here

httpbakery/agent/agent_test.go
```diff
@@ -461,9 +461,8 @@ func (c idmClient) IdentityFromContext(ctx context.Context) (bakery.Identity, []
 }
 
 func identityCaveats(dischargerURL string) []checkers.Caveat {
-	return []checkers.Caveat{{
-		Location:  dischargerURL,
-		Condition: "test condition",
+	return []checkers.Caveat{{Location: dischargerURL,
```
mhilton, line 464: and here

rogpeppe:
This is ancient and it's really superceded by the fact that third party caveat information can now be stored externally, so size isn't such a big concern any more.

PR #145, opened 2017-01-20
bakery: revert declaration simplification

rogpeppe:
After some thought, it has become clear that the path
ahead isn't that obvious and the declarations aren't so
easily simplified because of security concerns related
to the declaration caveat name changes. This reverts back
to the previous state of affairs until we can decide the
best way forward.

This reverts PR #139.


mhilton (approved):
LGTM
As this reverts the repository to a previous state I didn't look too closely at it, but the green bits seems sane.

fabricematrat (approved):

PR #147, opened 2017-01-26
bakery/agent: allow multiple users for a URL

rogpeppe:
This allows a serialized agent file to contain several agents for a
given third party.  Only the most recently added will be used, but
users can manually edit the serialized form to change the preference,
and a future change may allow the preference to be expressed otherwise.

With this in place, the put-agent command will be able
to create a new agent entry without worrying about
overwriting potentially irretrievable private key data.


mhilton:
OK seems reasonable

PR #148, opened 2017-01-31
Uses mgosession in mgostorage.

alesstimec:


rogpeppe (changes requested):
As discussed online, I don't think this is the right approach. Just create the with a new session when you need to. Asking for a session on every operation isn't a great pattern in general, as it can lose sequential consistency when mgo is not in the strongest consistency mode.

PR #152, opened 2017-03-13
bakery: add context to DeclaredIdentity call

rogpeppe:
This allows an IdentityClient to make context-specific
decisions about whether a given identity is allowed or not.


hatched (approved):
Thanks for all the context!

fabricematrat (approved):

PR #153, opened 2017-06-16
bakery/example: fix for new API

rogpeppe:
We fix the example so it works and its structure is more idiomatic.

We also clean up some of the error-related functions in
the httpbakery API, removing some deprecated
functions, shortening some names, and exporting
the WriteError function.


mhilton (approved):
LGTM

jrwren (approved):

PR #155, opened 2017-08-14
httpbakery: rework interaction-required protocol

rogpeppe:
We make the interaction-required error return all the
supported interaction methods directly in the response
rather than requiring the client to fetch the visit-URL
with "Accept: application/json".

All interaction methods now complete by producing
a "discharge token" which can then be presented to
the discharge endpoint to acquire the discharge macaroon.
It is the responsibility of the server to check that the
discharge token is acceptable for the provided third
party caveat.

There is no longer a requirement that all methods fit
the "visit/wait" protocol, which means that the
protocol is now capable of supporting a redirect-based
flow.

We split the bakerytest functionality up into two parts.
One is the Discharger functionality which many external
packages use; the other (Rendezvous) makes it easier
to test visit/wait-style interaction-required protocols.
The actual HTTP endpoints are left up to individual
clients to implement, as they vary a lot depending
on what is being tests.


bakerytest/rendezvous.go
```diff
@@ -0,0 +1,165 @@
...
+	r.waiting[dischargeId] = &dischargeFuture{
+		info: cav,
+		done: make(chan struct{}),
+	}
+	return dischargeId
+}
+
+// Info returns information on the given discharge id
+// and reports whether the information has been found.
+// It panics if there is no information found.
```
mhilton: Does it? it doesn't look like it does.

httpbakery/agent/agent.go
```diff
@@ -98,134 +38,138 @@ type Agent struct {
...
+	// LoginURL holds the URL from which to acquire
+	// a macaroon that can be used to complete the agent
+	// login. To acquire the macaroon, make a POST
+	// request to the URL with user and public-key
+	// parameters.
+	LoginURL string `json:"login-url"`
+}
+
+// interactor is a httpbakery.Interactor that performs interaction using the
 // agent login protocol. A Visitor may be encoded as JSON or YAML
```
mhilton: I suspect the part of the comment about a Visitor should be moved somewhere else.

httpbakery/agent/agent.go
```diff
@@ -24,72 +24,12 @@ import (
 
 var logger = loggo.GetLogger("httpbakery.agent")
 
-/*
-PROTOCOL
-
-An agent login works as follows:
-
-	    Agent                            Login Service
```
mhilton, line 32: it's a shame you haven't replaced this with a version for the new protocol, it would give a nice overview.

httpbakery/agent/agent.go
```diff
@@ -98,134 +38,138 @@ type Agent struct {
...
+	var p InteractionInfo
+	err := interactionRequiredErr.InteractionMethod("agent", &p)
+	if err != nil {
+		return nil, errgo.Mask(err)
+	}
+	if p.LoginURL == "" {
+		return nil, errgo.Newf("no login-url field found in agent interaction method")
 	}
-	u, err := url.Parse(a.URL)
+	logger.Infof("got agent interaction response; LoginURL=%q", p.LoginURL)
```
mhilton: Debugf surely

httpbakery/agent/agent_test.go
```diff
@@ -1,534 +1,189 @@
...
-	if err != nil {
-		httprequest.WriteJSON(w, http.StatusInternalServerError, httpbakery.Error{
-			Message: fmt.Sprintf("cannot create macaroon: %s", err),
-		})
-		return
-	}
-	err = httpbakery.NewDischargeRequiredError(m, "", authErr, req)
-	httpbakery.WriteError(ctx, w, err)
-}
+//
```
mhilton: please put a TODO at the top of this (or spomething along those lines)

httpbakery/agent/legacy_test.go
```diff
@@ -0,0 +1,341 @@
...
+	)
+	c.Assert(err, gc.IsNil)
+	ms, err := client.DischargeAll(context.Background(), m)
+	c.Assert(err, gc.IsNil)
+	authInfo, err := s.serverBakery.Checker.Auth(ms).Allow(context.Background(), bakery.LoginOp)
+	c.Assert(err, gc.IsNil)
+	c.Assert(authInfo.Identity, gc.Equals, bakery.SimpleIdentity("test-user"))
+}
+
+//
```
mhilton: please add a TODO/delete as is actually required here.

httpbakery/agent/protocol.go
```diff
@@ -0,0 +1,76 @@
...
+	POST /discharge
+L->A
+	Interaction-required error containing
+	an entry for "agent" with field
+	"login-url" holding URL $loginURL.
+A->L
+	GET $loginURL?username=$user&public-key=$pubkey
+	where $user is the username to log in as
+	and $pubkey is the public key of that username,
+	base-64 encoded (standard encoding)
```
mhilton: URL encoding please, it is in a URL after all? either that or head off more comments like this with an explanation of why use standard encoding (and be clear that it's URL escaped after encoding)

httpbakery/agent/protocol.go
```diff
@@ -0,0 +1,76 @@
...
+L->A
+	JSON response:
+		macaroon:
+			macaroon with "local" third-party-caveat
+			addressed to $pubkey.
+A->L
+	POST /discharge?token=self-discharged macaroon
```
mhilton: given it's likely size can we document that this can be in the post body in the same format too?

httpbakery/agent/protocol.go
```diff
@@ -0,0 +1,76 @@
...
+	The macaroon is binary-encoded, then base64
+	encoded with raw URL encoding.
```
mhilton: So we use the other base64 encoding here just to keep everyone on their toes?

httpbakery/browser.go
```diff
@@ -26,16 +32,116 @@ func OpenWebBrowser(url *url.URL) error {
...
+	return WebBrowserInteractionKind
+}
+
+// SetInteraction sets interaction information on the given error,
+// which should be an interaction-required error that's about
+// to be returned from a discharge request.
+//
+// The visitURL parameter holds a URL that should be
+// visited by the user in a web browser; the waitTokenURL parameter
+// holds a URL that can be long-polled to acquire the resulting
```
mhilton: probably not for this PR, but I think we should add a maximum timeout to the wait endpoints in general after which we return "still waiting" or something.

httpbakery/browser.go
```diff
@@ -26,16 +32,116 @@ func OpenWebBrowser(url *url.URL) error {
...
-		return ErrMethodNotSupported
+// Interact implements Interactor.Interact by opening a new web page.
+func (wi WebBrowserInteractor) Interact(ctx context.Context, client *Client, location string, irErr *Error) (*DischargeToken, error) {
+	var p webBrowserParams
+	if err := irErr.InteractionMethod(wi.Kind(), &p); err != nil {
+		logger.Infof("WebBrowserInteractor.Interact; no method %v", err)
+		return nil, errgo.Mask(err, errgo.Is(ErrInteractionMethodNotFound))
 	}
-	return OpenWebBrowser(u)
+	logger.Infof("WebBrowserInteractor.Interact %#v", p)
```
mhilton: again infof seems a little high here.

httpbakery/client.go
```diff
@@ -199,10 +222,10 @@ func (c *Client) Do(req *http.Request) (*http.Response, error) {
...
 // with a *DischargeError cause will be returned.
 //
-// If interaction is required by the user, the client's WebPageVisitor,
-// or VisitWebPage function will be used to perform interaction. An error
+// If interaction is required by the user, the client's InteractionMethods
+// will be used to perform interaction. An error
 // with a *InteractionError cause will be returned if this interaction
-// fails. See WebBrowserVisitor for a possible implementation of
+// fails. See WebBrowserWindowInteractor for a possible implementation of
 // WebPageVisitor.
```
mhilton: an InteractionMethod?

httpbakery/client.go
```diff
@@ -456,65 +466,168 @@ func (c *Client) AcquireDischarge(ctx context.Context, cav macaroon.Caveat, payl
 	if cause.Info == nil {
 		return nil, errgo.Notef(err, "interaction-required response with no info")
 	}
+	logger.Infof("got interaction-required error, Info %s", jsonMarshal(cause))
```
mhilton: In general perhaps consider the logging levels.

httpbakery/client.go
```diff
@@ -456,65 +466,168 @@ func (c *Client) AcquireDischarge(ctx context.Context, cav macaroon.Caveat, payl
...
+	// the token acquired by the interaction method.
+	m, err = c.acquireDischarge(ctx, cav, payload, token)
 	if err != nil {
 		return nil, errgo.Mask(err, IsDischargeError, IsInteractionError)
 	}
 	return m, nil
 }
 
+func jsonMarshal(x interface{}) string {
+	data, _ := json.Marshal(x)
```
mhilton: interesting, why is it safe to ignore the error here. I'm assuming that if there is some problem it just returns a zero string which is what you'd pust somewhere anyway then please could you add a doc-comment to that effect.

httpbakery/client.go
```diff
@@ -456,65 +466,168 @@ func (c *Client) AcquireDischarge(ctx context.Context, cav macaroon.Caveat, payl
...
 	}
-	waitResp, err := ctxhttp.Get(ctx, c.Client, waitURL.String())
+	return nil, &InteractionError{
+		Reason: errgo.Newf("no methods supported"),
+	}
+}
+
+// waitForMacaroon returns a macaroon from a legacy wait endpoint.
+func waitForMacaroon(ctx context.Context, client *Client, waitURL *url.URL) (*bakery.Macaroon, error) {
+
```
mhilton: This spare line probably isn't needed.

httpbakery/client_test.go
```diff
@@ -813,10 +725,23 @@ func (s *ClientSuite) TestDischargeWithInteractionRequiredErrorAndWebPageVisitor
 	c.Assert(httpbakery.IsInteractionError(errgo.Cause(err)), gc.Equals, true)
 	ierr, ok := errgo.Cause(err).(*httpbakery.InteractionError)
 	c.Assert(ok, gc.Equals, true)
-	c.Assert(ierr.Reason, gc.Equals, errCannotVisit)
+	c.Assert(errgo.Cause(ierr.Reason), gc.Equals, errCannotVisit)
 	c.Assert(resp, gc.IsNil)
 }
 
+//
```
mhilton: perhaps a TODO here?

httpbakery/error.go
```diff
@@ -82,20 +84,68 @@ type ErrorInfo struct {
...
 	// to acquire the discharge macaroon. A GET on
 	// this URL will block until the client has authenticated,
 	// and then it will return the discharge macaroon.
+	// This is deprecated - it is superceded by the InteractionMethods
+	// field.
 	WaitURL string `json:",omitempty"`
 }
 
+// SetInteraction sets the information for a particular
+// interaction kind to x.. The error should be an interaction-required
```
mhilton: s/\.\././

mhilton (commented):
A few comments from a first pass look, mostly superficial stuff while I work out what's going on.

httpbakery/discharge.go
```diff
@@ -273,3 +291,28 @@ func randomBytes(n int) ([]byte, error) {
 	}
 	return b, nil
 }
+
+// maybeBase64Encode encodes b as is if it's
+// OK to be passed as a URL form parameter,
+// or encoded as base64 otherwise.
+func maybeBase64Encode(b []byte) (s, s64 string) {
+	if utf8.Valid(b) {
```
mhilton, line 288: What about being utf-8 makes it OK to be passed as a form parameter? parts of it will still potentially be byte encoded.

httpbakery/browser.go
```diff
@@ -26,16 +32,116 @@ func OpenWebBrowser(url *url.URL) error {
...
+// The visitURL parameter holds a URL that should be
+// visited by the user in a web browser; the waitTokenURL parameter
+// holds a URL that can be long-polled to acquire the resulting
+// discharge token.
+//
+// The waitURL parameter is for legacy clients that don't
+// understand about discharge tokens and should
+// hold a URL that can be long-polled to acquire the
+// discharge macaroon.
+func (wi WebBrowserInteractor) SetInteraction(e *Error, visitURL, waitTokenURL, waitURL string) {
```
mhilton: This method seems backwards. If I'm a server that wants to allow the WebBrowser interaction, why would I want to create one just to add the URLs? Why not just make this a function it doesn't seem to use the receiver at all.

httpbakery/browser.go
```diff
@@ -26,16 +32,116 @@ func OpenWebBrowser(url *url.URL) error {
...
+		open = OpenWebBrowser
+	}
+	if err := open(visitURL); err != nil {
+		return nil, errgo.Mask(err)
+	}
+	return waitForToken(ctx, client, waitTokenURL)
+}
+
+// waitForToken returns a token from a the waitToken URL
+func waitForToken(ctx context.Context, client *Client, waitTokenURL *url.URL) (*DischargeToken, error) {
```
mhilton, line 153: As it's possible to do so. Let's support short polling with a timeout here. This will avoid errors where apache/haproxy decide to time out while we're still waiting.

httpbakery/error.go
```diff
@@ -82,20 +84,68 @@ type ErrorInfo struct {
...
+// error. This method will panic if x cannot be JSON-marshaled.
+// In general, this method should not be used directly - higher
```
mhilton: Why shouldn't I use this? when is it safe to use it? Does that imply that all interaction methods have to be supported in the bakery?

mhilton (commented):
In general this looks really good. I feel the server end of of how this will work hasn't been fully thought through though. I'd like to chat about it before giving a LGTM.

mhilton (approved):
LGTM with changes as discussed, at least to then iterate on later.

bakerytest/rendezvous.go
rogpeppe: Fixed comment.

httpbakery/agent/agent.go
rogpeppe: It's in protocol.go now.

httpbakery/agent/agent.go
rogpeppe: removed

httpbakery/agent/protocol.go
rogpeppe: Done, but does it make that much of a difference?  Macaroons need to be able to fit into headers anyway, and this one isn't likely to be that large (the one in the test came in at 332 bytes after URL encoding)

httpbakery/error.go
```diff
@@ -136,6 +186,8 @@ func ErrorToResponse(ctx context.Context, err error) (int, interface{}) {
 			panic(fmt.Sprintf("out of range version number %v", errorBody.version))
 		}
 	}
+	data, _ := json.Marshal(body)
```
cmars: Check the error.

httpbakery/error.go
```diff
@@ -136,6 +186,8 @@ func ErrorToResponse(ctx context.Context, err error) (int, interface{}) {
...
+	logger.Infof("ErrorToResponse -> %s", data)
```
cmars: This seems like trace-level log content.

httpbakery/error.go
```diff
@@ -161,6 +213,7 @@ func errorResponseBody(err error) *Error {
 		// error message but copy everything else.
 		errResp = *cause
 		errResp.Message = err.Error()
+		logger.Infof("it's a *Error %p; origInfo %#v; Info %#v", cause, cause.Info, errResp.Info)
```
cmars: logger.Tracef

httpbakery/discharge.go
cmars: Why not just base64-encode all tokens? Tokens are for machines, not humans.

httpbakery/discharge.go
```diff
@@ -183,31 +194,41 @@ type dischargeResponse struct {
...
-	} else {
-		id = []byte(r.Id)
+	id, err := maybeBase64Decode(r.Id, r.Id64)
+	if err != nil {
+		return nil, errgo.Notef(err, "bad caveat id")
 	}
 	var caveat []byte
 	if r.Caveat != "" {
+		// Note that it's important that when r.Caveat is empty,
+		// DischargeParams.Caveat is nil.
```
cmars: Assert this constraint?

httpbakery/client.go
```diff
@@ -456,65 +459,168 @@ func (c *Client) AcquireDischarge(ctx context.Context, cav macaroon.Caveat, payl
...
 		}
-		return nil, errgo.NoteMask(dischargeErr, "failed to acquire macaroon after waiting", errgo.Any)
+		return nil, errgo.NoteMask(err, "failed to acquire macaroon after waiting", errgo.Any)
 	}
 	var resp WaitResponse
-	if err := json.NewDecoder(waitResp.Body).Decode(&resp); err != nil {
+	if err := httprequest.UnmarshalJSONResponse(httpResp, &resp); err != nil {
 		return nil, errgo.Notef(err, "cannot unmarshal wait response")
 	}
-	if resp.Macaroon == nil {
```
cmars, line 515: What are the implications for legacy interaction?

httpbakery/client.go
cmars: This should return the error or panic.

httpbakery/client.go
cmars: logger.Tracef

httpbakery/client.go
```diff
@@ -456,65 +459,168 @@ func (c *Client) AcquireDischarge(ctx context.Context, cav macaroon.Caveat, payl
...
 // interact gathers a macaroon by directing the user to interact with a
-// web page.
-func (c *Client) interact(ctx context.Context, location, visitURLStr, waitURLStr string) (*bakery.Macaroon, error) {
-	visitURL, err := relativeURL(location, visitURLStr)
-	if err != nil {
-		return nil, errgo.Notef(err, "cannot make relative visit URL")
+// web page. The irErr argument holds the interaction-required
+// error response.
+func (c *Client) interact(ctx context.Context, location string, irErr *Error, payload []byte) (*DischargeToken, *bakery.Macaroon, error) {
+	mj, _ := json.Marshal(irErr.Info.InteractionMethods)
```
cmars: Check the error.

httpbakery/discharge.go
```diff
@@ -98,6 +103,9 @@ type DischargerParams struct {
 //		id: all-UTF-8 third party caveat id
 //		id64: non-padded URL-base64 encoded caveat id
 //		macaroon-id: (optional) id to give to discharge macaroon (defaults to id)
+//		token: (optional) value of discharge token
```
cmars, line 101: If we just base64-encoded, we could drop a field?

cmars (commented):
Let's be sure we're not losing test coverage, there's several commented out / removed tests in this PR -- though that may be WIP. Definitely want coverage asserting that legacy interaction still works as before.

bakerytest/rendezvous.go
rogpeppe: (it doesn't panic if there's no information found)

httpbakery/agent/protocol.go
rogpeppe: I've changed everything to accept all possible base64 encodings.

httpbakery/browser.go
rogpeppe: As discussed, I'm not sure. I definitely think we should have a loop here, but I'm not convinced whether the timeout parameter would help - what if the timeout is longer than the intermediate proxy's timeout?

httpbakery/client.go
rogpeppe: Removed this debugging remnant.

httpbakery/client.go
rogpeppe: It's only there for the debugging remnant log message which I've removed.

httpbakery/error.go
rogpeppe: removed debugging remnant (and the unchecked marshal)

httpbakery/error.go
rogpeppe: removed

httpbakery/browser.go
rogpeppe: Yes, but not in this PR.

httpbakery/client.go
rogpeppe: I've removed it.

httpbakery/client.go
rogpeppe: done

httpbakery/client.go
rogpeppe: removed

httpbakery/client.go
rogpeppe: removed

httpbakery/agent/agent.go
rogpeppe: done

httpbakery/agent/protocol.go
rogpeppe: We now accept any base64 encoding.

httpbakery/browser.go
rogpeppe: I've factored out this method into a function.

httpbakery/browser.go
rogpeppe: removed

httpbakery/discharge.go
rogpeppe: If it's valid ascii, it seems unnecessary to encode it as base64 - it takes more space as isn't as easy to debug.
I've chosen the same approach used by the macaroon encoding here.

I've also made the criteria a little more strict - it can't include ASCII control characters now.

httpbakery/error.go
rogpeppe: done

httpbakery/error.go
rogpeppe: Adjusted doc comment.

httpbakery/client.go
rogpeppe: We still support legacy interaction.

httpbakery/discharge.go
rogpeppe: It's something we're ensuring with the if above. I've rephrased the comment to hopefully make that clearer.

mhilton (approved):
LGTM

PR #157, opened 2017-09-14
httpbakery: allow base64-encoded token in WaitTokenResponse

rogpeppe:
This makes it easy for a visit-wait implementation to use a binary token.


mhilton (approved):
LGTM thanks very much

PR #158, opened 2017-09-21
bakery: ACLAuthorizer fine grained allow-public support

rogpeppe:
Some services wish to allow public access for some operations
but not for others. This change changes the GetACL method
so that it reports whether a given operation should allow
public access or not rather than having it as a global
setting for a given Authorizer.


mhilton (approved):
LGTM thanks

fabricematrat (approved):

PR #160, opened 2017-10-02
bakery: add Slice type

rogpeppe:
This makes it easier for a client to store undischarged macaroons,
which are more versatile because caveats can be added and discharges
obtained independently of the primary macaroon.

### DIFF
--- a/bakery/slice.go
+++ b/bakery/slice.go
@@ -1,0 +1,133 @@
+package bakery
+
+import (
+	"fmt"
+	"golang.org/x/net/context"
+	"time"
+
+	errgo "gopkg.in/errgo.v1"
+	"gopkg.in/macaroon-bakery.v2-unstable/bakery/checkers"
+	macaroon "gopkg.in/macaroon.v2-unstable"
+)
+
+// Slice holds a slice of unbound macaroons.
+type Slice []*Macaroon
+
+// Bind prepares the macaroon slice for use in a request. This must be
+// done before presenting the macaroons to a service for use as
+// authorization tokens. The result will only be valid
+// if s contains discharge macaroons for all third party
+// caveats.
+//
+// All the macaroons in the returned slice will be copies
+// of this in s, not references.
+func (s Slice) Bind() macaroon.Slice {
+	if len(s) == 0 {
+		return nil
+	}
+	ms := make(macaroon.Slice, len(s))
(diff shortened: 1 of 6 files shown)


bakery/slice.go
```diff
@@ -0,0 +1,127 @@
+package bakery
+
+import (
+	"context"
```
mhilton: wrong context package?

fabricematrat (approved):

bakery/slice.go
```diff
@@ -0,0 +1,127 @@
...
+		m1 := m.M().Clone()
+		m1.Bind(rootSig)
+		ms[i+1] = m1
+	}
+	return ms
+}
+
+// Purge returns a new slice holding all macaroons in s
+// that expire after the given time.
+func (ms Slice) Purge(t time.Time) Slice {
```
mhilton, line 41: This should probably be a little more intelligent. if Slice[0] has expired and gets purged doesn't it make the Slice somewhat meaningless, should it perhaps throw everything away then?

httpbakery/client.go
```diff
@@ -199,6 +199,12 @@ func (c *Client) DischargeAll(ctx context.Context, m *bakery.Macaroon) (macaroon
 	return bakery.DischargeAllWithKey(ctx, m, c.AcquireDischarge, c.Key)
 }
 
+// DischargeAllUnbound is like DischargeAll except that it does not
```
mhilton, line 202: does this function really add value, why not rely just using bakery.Slice.DischargeAll?

mhilton (approved):
LGTM with a couple of thoughts.

bakery/slice.go
rogpeppe: Excellent point - will do.

httpbakery/client.go
rogpeppe: It saves the client knowing about AcquireDischarge and Key, so I *think* it's just about worth it. I added it because I wanted to use it.

rogpeppe: Also, if DischargeAllUnbound isn't worth it, then DischargeAll probably isn't either. But I think it is... just about.

bakery/slice.go
rogpeppe: Done

PR #162, opened 2017-10-11
bakery: concurrency-safe ThirdPartyLocator

rogpeppe:
In general, ThirdPartyLocator implementations must be
concurrency safe, so document them as such and
fix bakery.ThirdPartyStore to work OK when its
methods are called concurrently.


mhilton (approved):
LGTM

bakery/keys.go
```diff
@@ -98,13 +99,15 @@ type ThirdPartyInfo struct {
...
 	// party at the given location. It returns ErrNotFound if no match is found.
+	// This method must be safe to call concurrently.
 	ThirdPartyInfo(ctx context.Context, loc string) (ThirdPartyInfo, error)
 }
 
 // ThirdPartyStore implements a simple ThirdPartyLocator.
 // A trailing slash on locations is ignored.
 type ThirdPartyStore struct {
-	m map[string]ThirdPartyInfo
+	mu sync.Mutex
```
alesstimec: you could use a RWMutex

bakery/keys.go
```diff
@@ -127,6 +133,8 @@ func canonicalLocation(loc string) string {
 
 // ThirdPartyInfo implements the ThirdPartyLocator interface.
 func (s *ThirdPartyStore) ThirdPartyInfo(ctx context.Context, loc string) (ThirdPartyInfo, error) {
+	s.mu.Lock()
```
alesstimec: and do a RLock, RUnlock here

alesstimec (approved):
LGTM with a comment

bakery/keys.go
rogpeppe: I usually don't consider RWMutexes unless the mutex is likely to be highly contended (e.g. the read or the write path is slow), but why not? Done.

PR #164, opened 2017-10-12
bakery: refactor AuthChecker.allowAny

rogpeppe:
The allowAny method was getting rather large. We factor it into
three methods, preparing the way for adding a way to let
operations authorize other operations.

This PR should not change any externally visible semantics.

### DIFF
--- a/bakery/checker.go
+++ b/bakery/checker.go
@@ -273,6 +273,8 @@ func (a *AuthChecker) Allowed(ctx context.Context) (*AuthInfo, map[Op]bool, erro
 		return a.newAuthInfo(used), nil, errgo.Mask(err)
 	}
 	ops := make(map[Op]bool)
+	// TODO this is non-deterministic; perhaps we should sort the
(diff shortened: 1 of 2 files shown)


bakery/checkers/operations.go
```diff
@@ -14,6 +14,8 @@ type opKey struct{}
 // given operations. An allow caveat will succeed only if one of the allowed
 // operations is in ops; a deny caveat will succeed only if none of the denied
 // operations are in ops.
+//
+// This uses the same context key as ContextWithOpAuthorizer.
```
mhilton: What's that then, and why are there two functions that do the same thing?

mhilton (approved):
LGTM with a thought

bakery/checkers/operations.go
rogpeppe: Ah, missed that - it's from the next PR. Ignore this comment - all should become clear in time.

PR #165, opened 2017-10-12
bakery: add OpsAuthorizer

rogpeppe:
This allows operations to be authorized via other
authorized operations, which has a number of possible
use cases.

### DIFF
--- a/bakery/authorizer.go
+++ b/bakery/authorizer.go
@@ -24,6 +24,17 @@ type Authorizer interface {
 	Authorize(ctx context.Context, id Identity, ops []Op) (allowed []bool, caveats []checkers.Caveat, err error)
 }
 
+// OpsAuthorizer is used to check whether an operation authorizes some other
+// operation. For example, a macaroon with an operation allowing general access to a service
+// might also grant access to a more specific operation.
+type OpsAuthorizer interface {
+	// AuthorizeOp reports which elements of queryOps are authorized by
+	// authorizedOp. On return, each element of the slice should represent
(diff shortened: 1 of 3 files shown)


bakery/authorizer.go
```diff
@@ -24,6 +24,17 @@ type Authorizer interface {
 	Authorize(ctx context.Context, id Identity, ops []Op) (allowed []bool, caveats []checkers.Caveat, err error)
 }
 
+// OpsAuthorizer is used to check whether an operation authorizes some other
+// operation. For example, an macaroon with an operation allowing general access to a service
```
mhilton: a macaroon

bakery/checkers/operations.go
```diff
@@ -14,6 +14,8 @@ type opKey struct{}
 // given operations. An allow caveat will succeed only if one of the allowed
 // operations is in ops; a deny caveat will succeed only if none of the denied
 // operations are in ops.
+//
+// This uses the same context key as ContextWithOpAuthorizer.
```
mhilton: HAven't I complained about this commend once?

mhilton (approved):
I think this is OK. Still don't really understand why ops are actually useful.

fabricematrat (approved):

bakery/checkers/operations.go
rogpeppe: Yeah, it snuck back in.

bakery/authorizer.go
rogpeppe: done

PR #166, opened 2017-10-16
httpbakery: update NewDischargeRequiredError

rogpeppe:
The current design stems from the fact that we needed
to keep the API compatible. Now that we have bakery.Macaroon,
we can tell the version from the macaroon that's being returned,
and we can also provide more useful defaults for the cookie
path and its name (the current default of creating a separate cookie
for every macaroon is not such a great idea).

This is an API-breaking change.

### DIFF
--- a/bakery/example/targetservice.go
+++ b/bakery/example/targetservice.go
@@ -112,8 +112,11 @@ func (srv *targetServiceHandler) writeError(ctx context.Context, w http.Response
 		return
 	}
 
-	herr := httpbakery.NewDischargeRequiredError(m, "/", derr, req)
-	herr.(*httpbakery.Error).Info.CookieNameSuffix = "auth"
+	herr := httpbakery.NewDischargeRequiredError(httpbakery.DischargeRequiredErrorParams{
+		Macaroon:      m,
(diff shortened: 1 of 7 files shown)


bakery/example/targetservice.go
```diff
@@ -112,7 +112,11 @@ func (srv *targetServiceHandler) writeError(ctx context.Context, w http.Response
 		return
 	}
 
-	herr := httpbakery.NewDischargeRequiredError(m, "/", derr, req)
+	herr := httpbakery.NewDischargeRequiredError(httpbakery.DischargeRequiredErrorParams{
+		Macaroon:      m,
+		OriginalError: derr,
+		Request:       req,
+	})
 	herr.(*httpbakery.Error).Info.CookieNameSuffix = "auth"
```
fabricematrat: use trhe cookie suffix name in the call above.

fabricematrat (approved):
LGTM

bakery/example/targetservice.go
rogpeppe: done

mhilton (approved):
LGTM

PR #167, opened 2017-10-18
httpbakery: add Oven type

rogpeppe:
This implements some functionality which almost every HTTP server
using the bakery will need - to translate from bakery.DischargeRequiredError
to the discharge-required error as returned from httpbakery.

### DIFF
--- a/httpbakery/oven.go
+++ b/httpbakery/oven.go
@@ -1,0 +1,84 @@
+package httpbakery
+
+import (
+	"net/http"
+	"time"
+
+	"golang.org/x/net/context"
+	errgo "gopkg.in/errgo.v1"
+
+	"gopkg.in/macaroon-bakery.v2-unstable/bakery"
+)
+
+// Oven is like bakery.Oven except it provides a method for
+// translating errors returned by bakery.AuthChecker into
+// errors suitable for passing to WriteError.
+type Oven struct {
+	// Oven holds the bakery Oven used to create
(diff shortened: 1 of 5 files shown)


httpbakery/oven.go
```diff
@@ -0,0 +1,84 @@
...
+	// Oven holds the bakery Oven used to create
+	// new macaroons to put in discharge-required errors.
+	*bakery.Oven
+
+	// AuthnExpiry holds the expiry time of macaroons that
+	// are created for authentication. As these are generally
+	// applicable to all endpoints in an API, this is usually
+	// longer than AuthzExpiry. If this is zero, DefaultAuthnExpiry
+	// will be used.
+	AuthnExpiry time.Duration
```
mhilton, line 26: I'm wondering if these should be Spelled AuthorizationExpiry and AuthenticationExpiry.

mhilton (approved):
LGTM

mhilton:
On further thoughts the authn authz stuff I find somewhat jarring. I'm not certain that the bakery should be having an opinion about what a token is for.

httpbakery/oven.go
rogpeppe: Maybe... I'm not sure I want to spell those names out everywhere though.

rogpeppe:
After discussion, we've decided to leave the authz stuff as is for the time being.

PR #169, opened 2017-10-20
bakery: remove multi-op support; rename MacaroonOpStore

rogpeppe:
With the advent of OpsAuthorizer, if there's a need to store
large numbers of operations in a macaroon, that's possible
outside of the Oven by allowing a "multi-" op to authorize
its component operations. This functionality is not used
by any existing service anyway, so delete it in the interests
of simplicity.

Because this functionality was the only reason that
Oven.NewMacaroon required an expiry time, we can
now remove that argument from the somwhat argument-overburdened
NewMacaroon method, but we'll do that in the next PR.

We also rename MacaroonOpStore to MacaroonVerifier
because MacaroonOpStore is a confusing name and
MacaroonVerifier better describes its role.
by using the OpsAuthorizer to authorize the


bakery/common_test.go
```diff
@@ -169,9 +169,5 @@ func basicAuthFromContext(ctx context.Context) (user, password string) {
 }
 
 func mustGenerateKey() *bakery.KeyPair {
```
mhilton, line 171: Do you really need this function at all now?

dependencies.tsv
```diff
@@ -1,6 +1,8 @@
+github.com/frankban/quicktest	git	fe41ed117766719e2d387ef4dfd9b12846de9da9	2017-09-21T13:10:42Z
```
mhilton, line 1: It'd be nice if you only needed one of gocheck and quicktest

mhilton (approved):
LGTM with a couple of thoughts

bakery/common_test.go
rogpeppe: mustGenerateKey is a bit more wieldy than bakery.MustGenerateKey in tests, so I decided to leave it around for now.

dependencies.tsv
rogpeppe: Will fix in a subsequent PR.

PR #170, opened 2017-10-20
bakery: remove expiry time from Oven.NewMacaroon

rogpeppe:
This is no longer needed, and was always a bit of a pain
and hard to explain, so remove it.


mhilton (approved):
LGTM

PR #171, opened 2017-10-20
internal/httputil: use gopkg.in/check.v1

rogpeppe:
This removes the quicktest dependency, as suggested by @mhilton.


mhilton (approved):
LGTM

PR #173, opened 2017-10-24
bakery: factor out identity checks

rogpeppe:
Identity checking via third party caveats and declaration caveats
is fragile and doesn't really fit as part of the core bakery
abstraction. This PR factors it out into another package
(bakery/identchecker) that provides an API almost identical
to the current bakery API, while leaving the underlying
operation checker functionality in the core bakery package.

In the process, it has become clear that the AllowAny primitive
is problematic so it's been removed.

Some API changes made in this PR:

- declaration caveats are treated as structural caveats, and
operate on the current macaroons being checked instead
of a map inferred beforehand. checkers.ContextWithMacaroons
is used to add the relevant context.

- support for allow and deny caveats is removed from
checkers, as these interacted unpleasantly with the operation
checker (and they didn't actually check full operations anyway,
so were misleading).

- LegacyMacaroonOp is added to bakery.OvenParams so that
it can know which operation to associate with legacy macaroons


bakery/checkers/declared.go
```diff
@@ -8,17 +8,26 @@ import (
...
+// macaroon slice and the name space to use to interpret caveats in
+// the macaroons.
+func ContextWithMacaroons(ctx context.Context, ns *Namespace, ms macaroon.Slice) context.Context {
+	return context.WithValue(ctx, macaroonsKey{}, macaroonsValue{
+		ns: ns,
+		ms: ms,
+	})
+}
+
+func MacaroonsFromContext(ctx context.Context) (*Namespace, macaroon.Slice) {
```
mhilton: Can we have a doc comment please?

bakery/identchecker/authorizer.go
```diff
@@ -65,17 +55,17 @@ func (openAuthorizer) Authorize(ctx context.Context, id Identity, ops []Op) (all
...
 // Authorize implements Authorizer.Authorize.
-func (closedAuthorizer) Authorize(ctx context.Context, id Identity, ops []Op) (allowed []bool, caveats []checkers.Caveat, err error) {
+func (closedAuthorizer) Authorize(ctx context.Context, id Identity, ops []bakery.Op) (allowed []bool, caveats []checkers.Caveat, err error) {
 	return make([]bool, len(ops)), nil, nil
 }
 
 // AuthorizerFunc implements a simplified version of Authorizer
 // that operates on a single operation at a time.
-type AuthorizerFunc func(ctx context.Context, id Identity, op Op) (bool, []checkers.Caveat, error)
+type AuthorizerFunc func(ctx context.Context, id Identity, op bakery.Op) (bool, []checkers.Caveat, error)
```
mhilton, line 64: Not certain about this, in general xFunc is an x interface that is implemented as a single function. It's a little odd here especially as it's nearly, but not quite the same interface. I had to look for quite some time to work out that Authorizer and AuthorizerFunc had slightly different interfaces. The question I was trying to answer was why does AuthorizerFunc.Authorize have more than one line.

bakery/identchecker/checker.go
```diff
@@ -0,0 +1,297 @@
...
+	// always have been obtained from a call to
+	// IdentityClient.DeclaredIdentity.
+	Authorizer Authorizer
+
+	// MacaroonVerifier is used to retrieve macaroon root keys
+	// and other associated information.
+	MacaroonVerifier bakery.MacaroonVerifier
+}
+
+func NewChecker(p CheckerParams) *Checker {
```
mhilton: doc comment please

bakery/identchecker/checker.go
```diff
@@ -0,0 +1,297 @@
...
+		OpsAuthorizer:    identityOpsAuthorizer{c},
+		MacaroonVerifier: p.MacaroonVerifier,
+	}
+	return &Checker{
+		checker: bakery.NewChecker(bp),
+		p:       p,
+	}
+}
+
+type Checker struct {
```
mhilton: doc comments on all of these please.

bakery/identchecker/checker.go
```diff
@@ -0,0 +1,297 @@
...
+
+// LoginOp represents a login (authentication) operation.
+// A macaroon that is associated with this operation generally
+// carries authentication information with it.
+var LoginOp = bakery.Op{
+	Entity: "login",
+	Action: "login",
+}
+
+type AuthChecker struct {
```
mhilton: doc comment please (and below)

bakery/identchecker/checker.go
```diff
@@ -0,0 +1,297 @@
...
+		ForAuthentication: true,
+	}
+}
+
+type identityOpsAuthorizer struct {
+	checker *Checker
+}
+
+func (a identityOpsAuthorizer) AuthorizeOps(ctx context.Context, authorizedOp bakery.Op, queryOps []bakery.Op) ([]bool, []checkers.Caveat, error) {
+	logger.Infof("identityOpsAuthorizer %#v %#v {", authorizedOp, queryOps)
```
mhilton: this looks like debug level at best, and below.

bakery/identchecker/checker.go
```diff
@@ -0,0 +1,297 @@
...
+	if identity == nil {
+		// Shouldn't happen if DeclaredIdentity behaves itself, but
+		// be defensive just in case.
+		return nil, errgo.Newf("no declared identity found in LoginOp macaroon")
+	}
+	c.identity_ = identity
+	return identity, nil
+}
+
+// inferIdentity ensures sure that we always use the same identity for authorization.
```
mhilton: s/ sure//

bakery/identchecker/checker.go
```diff
@@ -0,0 +1,297 @@
...
+	if len(caveats) != 0 {
+		return nil, caveats, nil
+	}
+	c.identity_ = identity
+	return identity, nil, nil
+}
+
+type identityKey struct{}
+
+type identityVal struct {
```
mhilton: Why do you need this type?

mhilton (approved):
LGTM with a few thoughts

bakery/checker.go
```diff
@@ -35,57 +27,82 @@ type Op struct {
 	Action string
 }
 
+// NoOp holds the empty operation, signifying no authorized
+// operation. This is always considered to be authorized.
+var NoOp = Op{}
```
alesstimec: i think an operation that is always considered un-authorized also makes sense...

alesstimec: a note like "see OpsAuthorizer for its use" would be in order...

bakery/checker.go
```diff
@@ -35,57 +27,82 @@ type Op struct {
...
+	// When not all operations can be authorized with the macaroons
+	// supplied to Checker.Auth, the checker will call AuthorizeOps
+	// with NoOp, because some operations might be authorized
+	// regardless of authority. NoOp will always be the last
+	// operation queried within any given Allow call.
+	AuthorizeOps(ctx context.Context, authorizedOp Op, queryOps []Op) ([]bool, []checkers.Caveat, error)
+}
+
 // AuthInfo information about an authorization decision.
 type AuthInfo struct {
```
alesstimec, line 76: hmmm.. i can see how i would want to pass a logger to BakeryParams and have all authorization decisions automatically logged at some logging level (say Info)...

bakery/checker.go
```diff
@@ -397,98 +345,81 @@ func (a *AuthChecker) allowAny(ctx context.Context, ops []Op) (authed, used []bo
...
+	}
+	// We've still got at least one operation unauthorized.
+	// Try to see if it can be authorized with no operation at all.
+	authedOK, caveats, err := a.checker.p.OpsAuthorizer.AuthorizeOps(ctx, NoOp, a.need)
 	if err != nil {
-		// TODO if there are macaroons supplied that have failed, perhaps we shouldn't
-		// do this but return those errors instead? Doing things the current
-		// way means that we lose the previous errors.
-		return nil, errgo.Notef(err, "cannot check permissions")
+		// TODO this probably means "can't check" rather than "authorization denied";
```
alesstimec: i'd return internal server error in this case....  i'd definitely want to monitor this, though..

bakery/checker.go
```diff
@@ -574,23 +463,18 @@ func (c *caveatSquasher) add0(cond string) bool {
...
-	case checkers.CondAllow,
-		checkers.CondDeny,
-		checkers.CondDeclared:
-		return false
+	if cond != checkers.CondTimeBefore {
+		return true
+	}
+	et, err := time.Parse(time.RFC3339Nano, args)
+	if err != nil || et.IsZero() {
+		// Again, if it doesn't seem valid, leave it alone.
```
alesstimec, line 467: return true if et.IsZero()?? hmmm.. i'd rather not.. it means something went wrong on our end..

bakery/identchecker/checker.go
alesstimec: godoc

bakery/identchecker/checker.go
alesstimec: godoc

bakery/identchecker/checker.go
```diff
@@ -0,0 +1,297 @@
...
+		p:       p,
+	}
+}
+
+type Checker struct {
+	checker *bakery.Checker
+	p       CheckerParams
+}
+
+func (c *Checker) Namespace() *checkers.Namespace {
```
alesstimec: godoc

bakery/identchecker/checker.go
```diff
@@ -0,0 +1,297 @@
...
+	return c.checker.Namespace()
+}
+
+func (c *Checker) Auth(mss ...macaroon.Slice) *AuthChecker {
```
alesstimec: godoc

bakery/identchecker/checker.go
```diff
@@ -0,0 +1,297 @@
...
+	errgo "gopkg.in/errgo.v1"
+	"gopkg.in/macaroon.v2-unstable"
+
+	"gopkg.in/macaroon-bakery.v2-unstable/bakery"
+	"gopkg.in/macaroon-bakery.v2-unstable/bakery/checkers"
+)
+
+var logger = loggo.GetLogger("bakery.identchecker")
+
+type CheckerParams struct {
```
alesstimec: godoc

bakery/identchecker/checker.go
alesstimec: godoc

bakery/identchecker/checker.go
```diff
@@ -0,0 +1,297 @@
...
+
+	// mu guards the identity_ field.
+	mu sync.Mutex
+
+	// identity_ holds the first identity discovered by AuthorizeOps
+	// so that all authorizations use a consistent identity.
+	identity_ Identity
+}
+
+func (c *AuthChecker) Allow(ctx context.Context, ops ...bakery.Op) (*AuthInfo, error) {
```
alesstimec: godoc

bakery/identchecker/checker.go
```diff
@@ -0,0 +1,297 @@
...
+		Caveats:           identityCaveats,
+		ForAuthentication: true,
+	}
+}
+
+type identityOpsAuthorizer struct {
+	checker *Checker
+}
+
+func (a identityOpsAuthorizer) AuthorizeOps(ctx context.Context, authorizedOp bakery.Op, queryOps []bakery.Op) ([]bool, []checkers.Caveat, error) {
```
alesstimec: godoc

alesstimec (commented):
Some notes from me..

bakery/checkers/declared.go
rogpeppe: done

bakery/identchecker/authorizer.go
rogpeppe: Yeah, it's a bit odd - it's essentially a non-bulk version of the same call for convenience. It's been like that a while and is used in a reasonable number of places. I think I'll leave it for now and add a comment if that's OK.

rogpeppe: Actually, the comment does say that it implements "a simplified version of Authorizer" so I think it's probably OK as is.

bakery/checker.go
rogpeppe: True in this case means the caveat will be added unchanged. That is, the "something wrong" will be preserved intact, which I think is the right thing to do here.

bakery/identchecker/checker.go
rogpeppe: done

bakery/identchecker/checker.go
rogpeppe: done

bakery/identchecker/checker.go
rogpeppe: done

bakery/identchecker/checker.go
rogpeppe: removed

bakery/identchecker/checker.go
rogpeppe: done

bakery/identchecker/checker.go
rogpeppe: I've added a comment.

bakery/checker.go

bakery/checker.go
rogpeppe: yeah, that would be nice. I think that's something that can be added later without backward incompatibility though.

bakery/checker.go
rogpeppe: changed to return an error (and updated the AuthorizeOps comment too)

bakery/identchecker/checker.go
rogpeppe: done

bakery/identchecker/checker.go
rogpeppe: done

bakery/identchecker/checker.go
rogpeppe: done

bakery/identchecker/checker.go
rogpeppe: done

bakery/identchecker/checker.go
rogpeppe: done

bakery/identchecker/checker.go
rogpeppe: done

bakery/identchecker/checker.go
rogpeppe: done

bakery/identchecker/checker.go
rogpeppe: deleted this method instead.

PR #176, opened 2017-10-26
bakery/identchecker: fix error messages

rogpeppe:
Make the error messages more similar to how they
were before to lessen migration hassles.


mhilton (approved):
LGTM

PR #181, opened 2017-11-02
Depend on versioned httprequest gopkg.in/juju/httprequest.v2

cmars:
Ease the migration path from bakery v1 to v2, which each depend on
mutually API-incompatible revisions of github.com/juju/httprequest.


alesstimec (approved):

rogpeppe:
Went with #182 instead.

PR #183, opened 2017-12-21
postgresrootkeystore: new package

rogpeppe:
This provides a root key store implementation for Postgres
similar to mgorootkeystore. Quite a bit of the logic is the
same, but enough is different that it doesn't quite seem
to justify factoring out the common code quite yet.


bakery/postgresrootkeystore/rootkey.go
```diff
@@ -0,0 +1,347 @@
...
+		created: now,
+		expires: now.Add(s.policy.ExpiryDuration + s.policy.GenerateInterval),
+		// TODO return just newId when we know we can always
+		// use non-text macaroon ids.
+		id:      []byte(fmt.Sprintf("%x", newId)),
+		rootKey: newKey,
+	}, nil
+}
+
+func randomBytes(n int) ([]byte, error) {
```
alesstimec, line 330: this could be moved to a common package as it is probably the same as in mgorootkeystore

alesstimec: same for afterEq and beforeEq

alesstimec (approved):

bakery/postgresrootkeystore/rootkey.go
rogpeppe: Too trivial to be worth it, I think. I'd like to factor out lots more code than that ideally :)

PR #190, opened 2018-01-10
httpbakery: do not duplicate cookies

rogpeppe:
When we retry a request, the net/http cookie logic appends
any existing cookies to the Cookie header, which means that
we can append the same cookie several times.


mhilton (approved):
LGTM

PR #191, opened 2018-01-24
httpbakery: be more careful about cookie expiry times

rogpeppe:
When there is clock skew between server and client, we
can receive a discharged macaroon that is apparently expired,
even though it's valid on the server.

We work around that issue by using a session cookie
(which will not be stored permanently) when the expiry
time is short.

We also use a long expiry time for the cookie when the
macaroon has no expiry time, rather than using a session
cookie as we were doing before.

### DIFF
--- a/httpbakery/client.go
+++ b/httpbakery/client.go
@@ -8,6 +8,7 @@ import (
 	"net/http/cookiejar"
 	"net/url"
 	"strings"
+	"time"
 
 	"github.com/juju/loggo"
 	"golang.org/x/net/context"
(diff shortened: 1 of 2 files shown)


httpbakery/client.go
```diff
@@ -410,7 +416,18 @@ func NewCookie(ns *checkers.Namespace, ms macaroon.Slice) (*http.Cookie, error)
...
+		// The macaroon doesn't expire - use a very long expiry
+		// time for the cookie.
+		expires = time.Now().Add(PermanentExpiryDuration)
+	} else if expires.Sub(time.Now()) < time.Minute {
+		// The macaroon might have expired already, or it's
+		// got a short duration, so treat it as a session cookie
+		// by setting Expires to the zero time.
+		expires = time.Time{}
+	}
+	cookie.Expires = expires
```
urosj, line 430: How is bakery treating this value?

urosj (commented):
Not entirely sure this is ok. We'll treat recently expired ones  as something that's OK and valid all the time and give it access. Which some other part of the system can then give even one without expiry ...

mhilton (approved):
LGTM

urosj (approved):
LGTM, this is client side only

PR #192, opened 2018-01-24
httpbakery/agent: support authinfo in environment variables

rogpeppe:
This makes it easy to get agent authinfo from environment
variables instead of a file.

### DIFF
--- a/httpbakery/agent/agent.go
+++ b/httpbakery/agent/agent.go
@@ -7,8 +7,11 @@
 package agent
 
 import (
+	"encoding/json"
 	"errors"
+	"io/ioutil"
 	"net/url"
+	"os"
 	"strings"
 
 	"github.com/juju/loggo"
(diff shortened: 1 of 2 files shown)


httpbakery/agent/agent.go
```diff
@@ -40,6 +41,63 @@ type Agent struct {
...
+// BAKERY_PRIVATE_KEY - the private key of the agent, base64 encoded.
+// BAKERY_PUBLIC_KEY - the public key of the agent, base64 encoded.
+// BAKERY_USERS - space-separated (username, URL) pairs of username
+// and identity manager URLs.
+//
+// For example:
+//
+// BAKERY_PUBLIC_KEY='wg/rUMm5cy9O00TaS9pgDIHnFMwhrisI0KDSg4+KkGM='
+// BAKERY_PRIVATE_KEY='S529MFrVUWB/VqcsPABJSAlnY/57fdHi4cL9HdK9ZQw='
+// BAKERY_USERS='https://api.jujucharms.com/identity bob https://api.staging.jujucharms.com/identity'
```
mhilton: This is significantly more complex than I had in mind. I was essentially looking for a fix to https://github.com/juju/charmstore/issues/776. This is not usable that simply. We need to be able to provide API access to jujucharms.com through a mechanism that it "set these two environment variables", "use the tools" type system. The same as people expect from AWS. I'm not interested that we have this crazy bakery setup inside the API that complexity should not be forced onto out users.

httpbakery/agent/agent_test.go
```diff
@@ -117,6 +121,36 @@ func (s *agentSuite) TestSetUpAuth(c *gc.C) {
...
+	_, err = f.Write(data)
+	c.Assert(err, gc.Equals, nil)
+	f.Close()
+
+	os.Setenv("BAKERY_AGENT_FILE", f.Name())
+
+	authInfo1, err := agent.AuthInfoFromEnvironment()
+	c.Assert(err, gc.Equals, nil)
+	c.Assert(authInfo1, jc.DeepEquals, authInfo)
+}
```
mhilton, line 152: Please also add a test for the error when the environment variable isn't set.

mhilton (approved):
LGTM modulo an additional test.

httpbakery/agent/agent_test.go
rogpeppe: done.

PR #193, opened 2018-01-25
Migrate to upstream mgo.

ExternalReality:


rogpeppe:
Please ensure tests pass before proposing, and that dependencies.tsv is up to date.

Also, I'm sure that this repo (which isn't really a juju package) should be dependent on github.com/juju/mgo. Why are we creating a new fork of mgo anyway and not using the community-supported fork directly, BTW?

jameinel:
We've always needed to keep at least 1 patch vs upstream for our releases. And even if we managed to get all of our patches upstream (and they do seem receptive) we still are likely to need a patch to fix a critical issue before a release. Depending on a repository where we don't have commit rights is dangerous.
Our intent is to follow globalsign/mgo almost verbatim. But *right* now there is a patch that changes serialization of time.Duration, and 2 or 3 patches that we're working for landing upstream.

(So mostly that if we need a quick release for a bugfix, we should be depending on something where we have commit rights)

rogpeppe:
Thanks, but I'm going to do this somewhat differentlly to preserve backward compatibility.

PR #194, opened 2018-01-25
Update import paths to upstream mgo traking fork.

ExternalReality:


rogpeppe:
Thanks, but I think I'll do this somewhat differently so that we can preserve backward compatibility, thanks.

PR #196, opened 2018-02-01
Update dependencies.tsv to match the mgo version used by Juju.

jameinel:
This brings in updated mgo, which seems to have a chance to give a
failure in the test suite around IPv6 vs IPv4 addresses. So I just
updated the test to follow.


dependencies.tsv
```diff
@@ -13,5 +13,5 @@ gopkg.in/check.v1	git	4f90aeace3a26ad7021961c297b22c42160c7b25	2016-01-05T16:49:
 gopkg.in/errgo.v1	git	66cb46252b94c1f3d65646f54ee8043ab38d766c	2015-10-07T15:31:57Z
 gopkg.in/juju/environschema.v1	git	7359fc7857abe2b11b5b3e23811a9c64cb6b01e0	2015-11-04T11:58:10Z
 gopkg.in/macaroon.v1	git	ab3940c6c16510a850e1c2dd628b919f0f3f1464	2015-01-21T11:42:31Z
-gopkg.in/mgo.v2	git	4d04138ffef2791c479c0c8bbffc30b34081b8d9	2015-10-26T16:34:53Z
+gopkg.in/mgo.v2	git	f2b6f6c918c452ad107eec89615f074e3bd80e33	2016-08-18T01:52:18Z
```
rogpeppe: Why not just update to latest mgo version? (3f83fa5005286a7fe593b055f0d7771a7dce4655)

rogpeppe (approved):
LGTM, but I don't particularly see that we should match the exact version used by Juju - when we update we generally update to the latest available version.

httpbakery/keyring_test.go
```diff
@@ -104,7 +104,7 @@ func (s *KeyringSuite) TestPublicKeyWrongURL(c *gc.C) {
 	client := httpbakery.NewHTTPClient()
 	_, err := httpbakery.PublicKeyForLocation(client, "http://localhost:0")
 	c.Assert(err, gc.ErrorMatches,
-		`cannot get public key from "http://localhost:0/publickey": Get http://localhost:0/publickey: dial tcp 127.0.0.1:0: .*connection refused`)
+		`cannot get public key from "http://localhost:0/publickey": Get http://localhost:0/publickey: dial tcp (127.0.0.1|\[::1\]):0: .*connection refused`)
```
rogpeppe, line 107: or just `dial tcp .*: ` to make us resilient against future similar changes?

jameinel: do you expect there will be a IPv7 anytime soon?
I wanted to keep the explict 127 or ::1 to be clear that we are talking about localhost.

dependencies.tsv
jameinel: done

httpbakery/keyring_test.go
jameinel: depends whether you want to care about the IP address. If all you care'd about was the "cannot get public key" then why have any of the trailing information?
`cannot get public key from "http://.*".*`

I don't really know what the goal of the test is. I'm happy to be prompted in a different direction.

jameinel:
This is intended to be a step along the path of moving away from gopkg.in/mgo.v2 entirely, but its easiest if we transition by having all the dependencies already passing the test suite on the existing mgo driver.

httpbakery/keyring_test.go
rogpeppe: We don't care about the details of the message, but it's nice to know that it happens because the connection was refused rather than some other random reason.

PR #197, opened 2018-02-01
Bring bakery.v2 to depend on the same version of mgo that Juju depends on

jameinel:
This runs the test suite cleanly with PGTESTDISABLE=1 go test ./...

I disabled the postgresql tests because there wasn't any instructions on how it wanted postgresql to be set up. And while I could install postgresql, it then complained about the user, and when I created a user, it then complained that password authentication didn't work.
However, the changes here shouldn't touch postgresql so I don't think it affects the test suite.


mhilton:
Closing this after landing #222

PR #198, opened 2018-02-02
use configurable logging

rogpeppe:
This means we can use an externally defined logger
if needed. There are a few places where this was awkward
because there was no place to put the Logger, but
those log messages don't seem important enough
to justify a backwardly incompatible API or separate
entry points, so they've been removed.

Rethinking the actual log messages printed is left
for another PR.

### DIFF
--- a/bakery/logger.go
+++ b/bakery/logger.go
@@ -1,0 +1,34 @@
+package bakery
+
+import (
+	"github.com/juju/loggo"
+	"golang.org/x/net/context"
+)
+
+// Logger is used by the bakery to log informational messages
+// about bakery operations.
+type Logger interface {
+	Infof(ctx context.Context, f string, args ...interface{})
+	Debugf(ctx context.Context, f string, args ...interface{})
+}
+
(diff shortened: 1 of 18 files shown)


bakery/logger.go
```diff
@@ -0,0 +1,32 @@
...
+// given logger name.
+func DefaultLogger(name string) Logger {
+	return loggoLogger{loggo.GetLogger(name)}
+}
+
+type loggoLogger struct {
+	logger loggo.Logger
+}
+
+func (l loggoLogger) Debugf(_ context.Context, f string, args ...interface{}) {
```
alesstimec: godoc

bakery/logger.go
```diff
@@ -0,0 +1,32 @@
...
+	l.logger.Debugf(f, args...)
+}
+
+func (l loggoLogger) Infof(_ context.Context, f string, args ...interface{}) {
```
alesstimec: godoc

alesstimec (approved):

mhilton (approved):
LGTM

PR #199, opened 2018-02-09
bakery/identchecker: add logger

rogpeppe:
We'd omitted the Logger parameter from identchecker.BakeryParams when adding the other logger stuff


mhilton (approved):
LGTM

PR #200, opened 2018-03-02
httpbakery: fix race when repeating requests

rogpeppe:
This bug ( https://bugs.launchpad.net/juju/+bug/1752203 )showed that there is a potential race condition when retrying an HTTP request. When there's a read or a write error on the request or response body, the request body can be closed asynchronously, and the Body field is accessed when doing so, so setting the Body field of a request after calling Client.Do is not a good idea.

We fix that by making a shallow copy of the Request for each bakery request, removing the need to modify the Body field.

I managed to verify that this kind of condition is a potential problem in a smaller program ( https://play.golang.org/p/xh8nkFuJhTm ) but despite trying quite hard, I was unable to come up with a test case using the bakery.

Nonetheless, this fix *should* fix the race...


mhilton (approved):
LGTM

PR #201, opened 2018-03-08
httpbakery: add global AllowInsecureThirdPartyLocator variable

rogpeppe:
This means that parallel tests can use bakerytest.NewDischarger without
worrying about global races involving http.DefaultTransport.


fabricematrat (approved):

mhilton (approved):
LGTM

PR #203, opened 2018-03-23
httpbakery: use OpenWebBrowser method in legacy interactor

rogpeppe:
It was an oversight that we didn't use the configured OpenWebBrowser field.

Also move the addCookie function out of the httpbakery/agent production code,
as it's not used any more, and comment that the LoginCookie function
is deprecated.


mhilton (approved):
LGTM

PR #205, opened 2018-04-23
bakery: add PrivateKey.Public method

rogpeppe:
It's easy to derive a public key from the private key, so provide
a method to do so, potentially allowing smaller private key tokens
if needed.


mhilton (approved):
LGTM

PR #206, opened 2018-05-22
httpbakery/agent: support MultiVisitor

rogpeppe:
We'd like to be able to use agent authentication along with
other methods, and the current implementation doesn't
allow that (that's fixed in bakery.v2, but many packages
can't use that yet), so we implement it.

The code uses the same external conventions as bakery.v2.

Note that this PR is targeted at v2-unstable-fork, not v2-unstable.


mhilton (approved):
LGTM

PR #210, opened 2019-04-08
bakery/checkers: add Namespace.Equal

rogpeppe:
This will allow us to use Namespace with cmp.Equal.

### DIFF
--- a/bakery/checkers/namespace.go
+++ b/bakery/checkers/namespace.go
@@ -18,6 +18,23 @@ type Namespace struct {
 	uriToPrefix map[string]string
 }
 
+// Equal reports whether ns2 encodes the same namespace
+// as the receiver.
+func (ns1 *Namespace) Equal(ns2 *Namespace) bool {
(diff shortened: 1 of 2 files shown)


bakery/checkers/namespace.go
```diff
@@ -18,6 +18,26 @@ type Namespace struct {
...
 }
 
+// Equal reports whether ns2 encodes the same namespace
+// as the receiver.
+func (ns1 *Namespace) Equal(ns2 *Namespace) bool {
+	if ns1 == ns2 {
+		return true
+	}
+	if ns1 == nil || ns2 == nil {
+		return (ns1 == nil) == (ns2 == nil)
```
mhilton: wouldn't `return false` suffice here? the True case will be caught above.

mhilton (approved):
LGTM with one thought.

frankban (approved):
Looks good thanks!

bakery/checkers/namespace.go
rogpeppe: good point. i folded it into the test above.

PR #211, opened 2019-04-08
bakery: use quicktest not gocheck

rogpeppe:
This is part of an ongoing effort to reduce dependencies.
To avoid one huge PR, we're doing this in stages.


frankban (approved):
Looks good!

mhilton (approved):
LGTM

PR #217, opened 2019-04-09
bakery: avoid direct loggo dependency

rogpeppe:
If anyone actually needs these log messages
sending to loggo, they can do so by implementing
bakery.Logger themselves.


mhilton (approved):
LGTM. I suppose you could log to the standard library's log package, but I think most of the time if you want logging you're better off implementing it yourself.

PR #219, opened 2019-12-11
Add a lock to the PostgreSQL store initialization script

alnvdl:
When stress testing an indirect user of this library, the following error was being returned:
```
cannot get macaroon: cannot initialize database: cannot initialize table: pq: tuple concurrently updated
```

One possible explanation for this is that `CREATE OR REPLACE FUNCTION` is [not really safe for being called concurrently](https://stackoverflow.com/questions/40525684/tuple-concurrently-updated-when-creating-functions-in-postgresql-pl-pgsql). If that happens, the store will be left forever in an error state.

This PR attempts to fix that by setting up a lock and a transaction in the initialization script so that only one concurrent store can execute it at a time.

### DIFF
--- a/bakery/postgresrootkeystore/sql.go
+++ b/bakery/postgresrootkeystore/sql.go
@@ -23,6 +23,15 @@ const (
 )
 
 var initStatements = `
+BEGIN;
+
+-- Set up an advisory lock so that only one thread can issue the statements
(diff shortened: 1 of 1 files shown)


mhilton (approved):
LGTM

PR #220, opened 2020-03-04
bakery/dbrootkeystore: add context to Backing.

mhilton:
Add an optional ContextBacking interface which may be implemented by
a Backing in order that a context.Context is passed to the Backing.

The bakery/mgorootkeystore has also been updated to implement the
new ContextBacking interface and provides a mechanism to attach a
*mgo.Session to a context, so that all requests are made with that
session.


bakery/mgorootkeystore/rootkey.go
```diff
@@ -132,9 +177,68 @@ func (b backing) FindLatestKey(createdAfter, expiresAfter, expiresBefore time.Ti
...
+func (b backing) runWithContext(ctx context.Context, f func(*mgo.Collection)) error {
+	s := sessionFromContext(ctx)
+	if s == nil {
+		s = b.coll.Database.Session
+	}
+	s = s.Clone()
+
+	c := make(chan struct{})
+
+	go func() {
```
alesstimec, line 216: is this run in a go routine just so we can check if the context is cancelled in the meantime? sure, runWithContext will return once the context is cancelled.. but the go routine will still be running.. do we care about that?

alesstimec (approved):
LGTM

PR #221, opened 2020-04-17
Check PKs in bakery example

tbg:
In the example, the target service was using an insecure third party
locator, so it wasn't really demonstrating much.


tbg:
Actually, I am realizing only now that absent a cache hit (i.e. prior to this PR) the auth service would be contacted and asked for its pub key (over http, which is what the `AllowInsecure()`) is for). So in a sense, it was authenticating, though via a shortcut. I still think this is much clearer now.

PR #222, opened 2021-03-02
Use github.com/juju/mgo/v2

wallyworld:
Update deps to use github.com/juju/mgo/v2 instead of gopkg.in/mgo.v2

Also fix a couple of tests to make them work with go 1.14


hpidcock (approved):

mhilton (approved):
LGTM

PR #227, opened 2021-03-08
bakery: use external macaroondb

mhilton:


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,49 @@
...
+    runs-on: ubuntu-latest
+    container:
+      image: ubuntu
+      volumes:
+      - /etc/ssl/certs:/etc/ssl/certs
+    services:
+      postgres:
+        image: ubuntu/postgres
+        env:
+          POSTGRES_PASSWORD: shei4eeJoo3Pee0rahng
```
cmars: I hope this was a password generated for this purpose and not reused from elsewhere...

To avoid confusion there I'd either use Github Secrets for this (https://docs.github.com/en/actions/reference/encrypted-secrets) or use something simple like "password" since it's a test throwaway database (like https://docs.github.com/en/actions/guides/creating-postgresql-service-containers).

cmars (approved):
LGTM with a note below on the postgres password.

.github/workflows/ci.yaml
mhilton: It certainly was generated for this purpose.

PR #236, opened 2021-03-09
Use github.com/go-macaroon-bakery/macaroon-bakery/v3 instead of gopkg.in/macaroon-bakery.v3

wallyworld:
Use a versioned go mod path instead of gopkg.in as there's now no need to be compatible with any non go mod callers.


mhilton (approved):
LGTM

PR #277, opened 2022-08-23
Update seekerFromBody to handle io.nopCloserWriterTo in Go 1.19

### LINKED ISSUE #276: Go 1.19: `request body is not seekable` error
```go
func test(url string, client *httpbakery.Client) {
	b := []byte{}
	req, _ := http.NewRequest("POST", url, bytes.NewReader(b))
	response, err := client.Do(req)
	fmt.Println(err)
}
```
This works fine on Go <= 1.18, but on Go 1.19, it is failing with the error
```
request body is not seekable
```

Looking into the code, it seems the issue is with the function `httpbakery/request.go:seekerFromBody`. In the Go 1.19 release notes:
> [NopCloser](https://tip.golang.org/pkg/io/#NopCloser)'s result now implements [WriterTo](https://tip.golang.org/pkg/io/#WriterTo) whenever its input does.

So the check [here](https://github.com/go-macaroon-bakery/macaroon-bakery/blob/v3/httpbakery/request.go#L176) is not quite correct - we need to compare it against the type of `io.nopCloserWriterTo`.
I wonder if there's actually a more robust way to do this.

barrettj12:
Fixes #276. I've added unit tests too.


httpbakery/request.go
```diff
@@ -3,12 +3,12 @@ package httpbakery
 import (
 	"context"
 	"io"
-	"io/ioutil"
 	"net/http"
 	"reflect"
 	"sync"
 	"sync/atomic"
 
+	"bytes"
```
benhoyt: Nit: move into stdlib section above "context".

httpbakery/request.go
```diff
@@ -159,7 +159,8 @@ func (r *readStopper) Close() error {
 	return r.rreq.close()
 }
 
-var nopCloserType = reflect.TypeOf(ioutil.NopCloser(nil))
+var nopCloserType = reflect.TypeOf(io.NopCloser(nil))
+var nopCloserWriterToType = reflect.TypeOf(io.NopCloser(bytes.NewReader([]byte{})))
```
benhoyt, line 163: Just for reference, I don't love the approach we're using here (see comments [here](https://github.com/go-macaroon-bakery/macaroon-bakery/issues/276)), but it seems a reasonable and minimally-invasive compromise.

httpbakery/request_test.go
```diff
@@ -0,0 +1,55 @@
+package httpbakery_test
+
+import (
+	"bytes"
+	qt "github.com/frankban/quicktest"
+	"github.com/go-macaroon-bakery/macaroon-bakery/v3/httpbakery"
+	"io"
+	"testing"
```
benhoyt: Nit: same comment about import grouping.

httpbakery/export_test.go
```diff
@@ -5,3 +5,5 @@ type PublicKeyResponse publicKeyResponse
 const MaxDischargeRetries = maxDischargeRetries
 
 var LegacyGetInteractionMethods = legacyGetInteractionMethods
+
+var SeekerFromBody = seekerFromBody
```
benhoyt: I can see why this is attractive, but is it difficult to use a real `http.Request` (maybe [`httptest.NewRequest`](https://pkg.go.dev/net/http/httptest#NewRequest)) so this test is more real, and to avoid testing internals?

barrettj12: Done

benhoyt (approved):

wallyworld (approved):

alesstimec (approved):

PR #301, opened 2024-07-10
feat(rootkeys): add new constructor for root key store

jack-w-shaw:
Previously, the only constructor we had for bakery.RootkeyStore took in a Backing, but if that backing also implements ContextBacking, we prefer that backing.

Changes being made in Juju mean tha we naturally wish to implement ContextBacking. Our db methods require a context. However, to make use of the existing NewStore constructor, we would need to implement stub methods without contexts.

We do not wish to expose these.

Add a new constructor which directly takes in a ContextBacking as it's backing. This means we will no longer need to implement stub context-less methods.

## QA Steps

In tests, all we wish to assert is that we can construct a functioning RootKeyStore without implementing `GetKey`, etc.

Add a unit test and a fake backing to test this.

```
go test ./bakery/dbrootkeystore/...
```


SimonRichardson (approved):
In the future, we might want to start dropping the Backing interface in favor of the ContextBacking.

kian99 (approved):

bakery/dbrootkeystore/rootkey_test.go
```diff
@@ -422,30 +422,66 @@ func TestContextBackingTakesPrecedence(t *testing.T) {
 	c.Assert(key1, qt.DeepEquals, key2)
 }
 
-type contextBacking struct {
+type fullContextBacking struct {
 	b dbrootkeystore.Backing
 }
 
-func (b contextBacking) GetKey(id []byte) (dbrootkeystore.RootKey, error) {
+func (b fullContextBacking) GetKey(id []byte) (dbrootkeystore.RootKey, error) {
```
alesstimec, line 429: perhaps a godoc explaining why this is returning an error for posterity..

alesstimec (approved):
LGTM, but naming is becoming a bit of a mess..

alesstimec:
> In the future, we might want to start dropping the Backing interface in favor of the ContextBacking.

Hopefully juju won’t be using macaroons much longer :)

